PR #3, opened 2015-02-17
Add basic fulfillment documentation.

dianakhuang:
@stephensanchez Any sort of comments/questions/things to expand upon here would be useful.


stephensanchez:
Looks good at a high level. I think the one thing we can elaborate on is return values, errors, and configuration of the relationship between Fulfillment Modules and Product Types. 

Regarding return values and errors, I imagined each of these fulfillment or revoke operations could return a JSON-serializable dict response of {'status':'success'} or {'status':'failure', 'errorType':'Connection Refused'}. 

The configuration of Fulfillment Modules should be something we can adjust on the fly as we build new Fulfillment Modules, so that can be a simple dictionary in the settings.py for the ecommerce service that does the appropriate mapping of ProductType to Class. 

If you're concerned that this is too granular to write now, and we'll end up re-writing it, I think some empty sections with TODO's will be fine to make sure we revisit the section after implementation.

dianakhuang:
That's a lot of great stuff to add in! I'll start working on putting it in. I think it's more fleshed out in your head than it is on the wiki page, so I definitely wanted to get your feedback early.

dianakhuang:
@rlucioni feedback on this would be much appreciated.

docs/fulfillment.rst
```diff
@@ -0,0 +1,50 @@
...
+    fulfill_product(product)
+
+    revoke_product(product)
+
+
+For each ``ProductType`` that we develop, there will be a corresponding module that will extend this interface and be responsible for fulfilling order items of that particular ``ProductType``. The mapping between each ``ProductType`` and the corresponding fulfillment module will be configured in ``settings.py`` as a simple dictionary.
+
+``fulfill_product`` will fulfill the given product (i.e. enroll a student in a course, upgrade the student to a verified certificate).
+
+``revoke_product`` will do the opposite (i.e. unenroll students from courses, downgrade people from veriied).
```
rlucioni: Nitpick, but should be `i.e.,`.

docs/fulfillment.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Both of these methods will return JSON-serializable dictionary containing both the status and any important status notes.
```
rlucioni: Should be "*a* JSON-serializable dictionary."

rlucioni:
Just some nitpicks, otherwise this sounds reasonable to me, :+1: . As @stephensanchez said, I think this will be worth revisiting once fulfillment is implemented so that we can add some more specifics.

stephensanchez:
@rlucioni absolutely.

stephensanchez:
Also :+1:  from my side.

PR #5, opened 2015-02-23
Add initial application overrides

stephensanchez:
Adding setup.py for project

Try to add package name to projects.

Add package names to paths.

Commit all transactions in data migrations


wedaly:
@stephensanchez Could you please separate this into two PRs: one that sets up the initial project and a second that includes just the data migration?

rlucioni:
@stephensanchez, the [cookiecutter-django](https://github.com/pydanny/cookiecutter-django/tree/master/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D) project doesn't use an `apps` directory, instead exposing apps directly under the main Django project directory/package. I think the motivation behind this layout is that dotted imports don't all need to have ".apps" in them.

The [edx-ecommerce](https://github.com/edx/edx-ecommerce/tree/master/ecommerce) project is laid out this way, and I'd like the layout of these two projects to be consistent. Would you object to removing the `apps` directory?

extensions/apps/catalogue/migrations/0001_initial.py
```diff
@@ -0,0 +1,276 @@
...
+                'abstract': False,
+            },
+            bases=(models.Model,),
+        ),
+        migrations.CreateModel(
+            name='Product',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('structure', models.CharField(default='standalone', max_length=10, verbose_name='Product structure', choices=[('standalone', 'Stand-alone product'), ('parent', 'Parent product'), ('child', 'Child product')])),
+                ('upc', oscar.models.fields.NullCharField(unique=True, verbose_name='UPC', max_length=64, help_text='Universal Product Code (UPC) is an identifier for a product which is not specific to a particular  supplier. Eg an ISBN for a book.')),
```
rlucioni: Nitpick, but the help text here should be "Universal Product Code (UPC) is an identifier for a product which is not specific to a particular supplier (e.g., an ISBN for a book)."

extensions/apps/order/models.py
```diff
@@ -0,0 +1 @@
+from oscar.apps.order.models import *  # noqa
```
rlucioni: Since we don't use Pyflakes, we can get rid of this `# noqa` comment. Might want to replace with with `# pragma: no cover` so it's not included in coverage reports.

extensions/apps/partner/admin.py
```diff
@@ -0,0 +1 @@
+from oscar.apps.partner.admin import *  # noqa
```
rlucioni: Same comment here about `# noqa`.

extensions/apps/partner/migrations/0003_auto_20150223_1130.py
```diff
@@ -0,0 +1,40 @@
...
+
+    """
+    # Create all our Product Types.
+    Partner = apps.get_model("partner", "Partner")
+    edx = Partner(code="edx", name="edX")
+    edx.save()
+
+    StockRecord = apps.get_model("partner", "StockRecord")
+    Product = apps.get_model("catalogue", "Product")
+    honor_seat = Product.objects.get(upc="000000000002")
```
rlucioni: We can't use UPC `000000000001`?

extensions/settings/edx_oscar.py
```diff
@@ -0,0 +1,45 @@
...
+    },
+}
+
+# TODO: Replace with new Authentication backend
+AUTHENTICATION_BACKENDS = (
+    'oscar.apps.customer.auth_backends.EmailBackend',
+    'django.contrib.auth.backends.ModelBackend',
+)
+
+OSCAR_DEFAULT_CURRENCY = 'USD'
```
rlucioni: There's a missing newline here.

extensions/apps/catalogue/migrations/0001_initial.py
stephensanchez: Hmm.. this was the original migration from Django Oscar core. I'd prefer to keep it the same as the initial project unless there's a bug, in which case we should fix the core then re-copy here, I'd think.

extensions/apps/order/models.py
stephensanchez: Good call.

extensions/apps/partner/migrations/0003_auto_20150223_1130.py
stephensanchez: We're using 000000000001 in the migration for the catalogue project, so this is the 'next' UPC

stephensanchez:
Regarding the "apps" directory, I think we can get rid of it. This is how django-oscar is laid out, so I kept the same package structure, but I think I'd rather be consistent with our other projects.

requirements/test.txt
```diff
@@ -0,0 +1,8 @@
+-r base.txt
+
+coverage==3.7.1
+ddt==0.8.0
+django-nose==1.3
+mock==1.0.1
+pep8==1.5.7
+pylint<1.0
```
rlucioni: Why `pylint<1.0`? I think this should be `pylint==1.4.1`; 1.4.1 is the latest release.

extensions/apps/catalogue/migrations/0001_initial.py
rlucioni: That's fair, this isn't too important anyway.

extensions/apps/partner/migrations/0003_auto_20150223_1130.py
rlucioni: Got it, thanks for the clarification.

extensions/settings/edx_oscar.py
stephensanchez: added a newline.

stephensanchez:
@wedaly Removing the south migrations cleared this up a bit. Are you OK with the one PR now, or do you still want this split up for logical separation of actions? (I squashed a bit early, can I still split it up?)

requirements/test.txt
rlucioni: Hmm, the pep8 version in use here is old too; latest is 1.6.2.

rlucioni: One more: ddt is now at version 1.0.0. They just keep coming!

Makefile
```diff
@@ -0,0 +1,66 @@
...
+generate_fake_translations: extract_translations dummy_translations compile_translations
+
+pull_translations:
+	tx pull -a
+
+update_translations: pull_translations generate_fake_translations
+
+static:
+	python ecommerce/manage.py collectstatic --noinput
+	python ecommerce/manage.py compress
```
rlucioni: Looks like this is leftover from the ecommerce Makefile; it should be using the `manage.py` file at `extensions/manage.py`.

rlucioni:
One last nitpick about a Makefile target. Otherwise, :+1: .

PR #6, opened 2015-02-24
Fix uniqueness constraint violation

rlucioni:
Also confines Oscar-specific imports to Oscar settings. @stephensanchez, could you take a look at this when you have a minute?


stephensanchez:
Looks good! :+1:

PR #9, opened 2015-02-24
Move settings back within extensions package

rlucioni:
This moves `settings` back within `extensions` so that services which install `extensions` can access Oscar settings.

@stephensanchez, I think we finally got it. Can you take a look at this when you're able?


stephensanchez:
:+1:

rlucioni:
Coverage isn't good, but tests are passing. We can fix coverage separately.

PR #10, opened 2015-02-24
Sanchez/fix data migration

stephensanchez:
Fixes the data migration parent/child relationship.

@rlucioni


extensions/order/processing.py
```diff
@@ -0,0 +1,29 @@
...
+                              line_quantities, **kwargs):
+        self.validate_shipping_event(
+            order, event_type, lines, line_quantities, **kwargs)
+
+        if event_type.name == 'Shipped':
+            print "Shipped order..."
+
+        shipping_event = self.create_shipping_event(
+            order, event_type, lines, line_quantities,
+            reference=kwargs.get('reference', None))
```
rlucioni: Missing newline here.

rlucioni:
Sick coverage, bro. Just one tiny nitpick, otherwise :+1: .

extensions/order/processing.py
stephensanchez: woops, this file wasn't supposed to be in this PR :smile:

PR #11, opened 2015-02-25
Fulfillment API

stephensanchez:
https://openedx.atlassian.net/browse/XCOM-108
This is the Fulfillment API framework. The implementation of the EnrollmentFulfillmentModule will be an additional PR on top of this. 

This API introduces a configuration that allows a single Fulfillment Module to handle multiple types of products, instead of a one-to-one relationship.  All products in an order are grouped by their product type, then all configured Fulfillment Modules are checked to see if they support any number of the product types. 

Any remaining products that are not supported by the configuration must be marked as a Fulfillment Error, as well as any server-side errors. This introduces a new set of "statuses" for the Order Lines, separate from the Orders themselves. This is now declared in _oscar.py settings.

When you have a chance, please review and let me know what you think.
@dianakhuang @rlucioni
CC: @wedaly @rock345


extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
+""" Fulfillment API for processing orders
+
+Common interface designed to allow the fulfillment of orders for any type of product introduced into our catalog.
+Calls down to a common interface for fulfillment of each line of an order based on a product type to determine if we
+can successfully fulfill the product. Success can be reported back based on each line item in the order.
+
+"""
+import logging
+from django.conf import settings
```
rlucioni: Nitpick, but could you separate the standard library import from the third-party imports with a newline?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
...
+from django.utils import importlib
+
+import errors
+
+log = logging.getLogger(__name__)
```
rlucioni: Nitpick, but convention seems to be to call this `logger`.

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
...
+        lines (List of Lines): A list of Line items in the Order that should be fulfilled.
+
+    Returns:
+        The modified Order and Lines. The status of the Order, or any given Line item, may be 'Complete', or
+        'Fulfillment Error' based on the result of the fulfillment attempt.
+
+    """
+    if 'Complete' not in order.available_statuses():
+        raise errors.IncorrectOrderStatusError(
+            "Order has a current status of [%s] which cannot be fulfilled.", order.status
```
rlucioni: Could you use `.format` to interpolate the order status?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Construct a dict of lines by their product type.
+    line_items = {}
+    for line in lines.all():
+        line_items.setdefault(line.product.product_class.name, []).append(line)
+
+    # Once all line items are organized by type, we can iterate over the Fulfillment Modules
+    # defined in our configuration and fulfill line items in the order they are designated by
+    # the configuration. Remaining line items should be marked with a fulfillment error since
+    # we have no configuration that allows them to be fulfilled.
```
rlucioni: In the event of a fulfillment error, should the entire order be rolled back? The prospect of handling partially fulfilled orders makes me uncomfortable, although Oscar does seem to include good support for it (referring to the line status pipeline).

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
...
+                for line in related_lines:
+                    line.set_status("Fulfillment Configuration Error")
+                    line.save()
+
+    # Check to see if any line items in the order have not been accounted for by a FulfillmentModule
+    # Any product does not line up with a module, we have to mark a fulfillment error.
+    for product_type in line_items.keys():
+        log.error("Product Type [%s] in order does not have an associated Fulfillment Module", product_type)
+        for line in line_items[product_type]:
+            line.set_status("Fulfillment Configuration Error")
```
rlucioni: Could you define a constant for this string (e.g., `FULFILLMENT_CONFIGURATION_ERROR_LINE_STATUS`) and use that instead of the raw string?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,97 @@
...
+            line.save()
+
+    # Check if all lines are successful, or there were errors, and set the status of the Order.
+    order_status = 'Complete'
+    for line in lines.all():
+        if line.status != 'Complete':
+            log.error('There was an error while fulfilling order [%s] for user [%s]', order.id, order.user.username)
+            order_status = 'Fulfillment Error'
```
rlucioni: I think these status strings ('Fulfillment Error' and 'Complete') could also benefit from being tied to constants.

extensions/fulfillment/tests/test_api.py
```diff
@@ -0,0 +1,101 @@
...
+from oscar.test import factories
+
+
+class FakeFulfillmentModule(FulfillmentModule):
+    """Fake Fulfillment Module used to test the API without specific implementations."""
+
+    def fulfill_product(self, order, lines):
+        """Fulfill product. Mark all lines success."""
+        for line in lines:
+            line.set_status('Complete')
```
rlucioni: Looks like we might want to pull these status strings out to the Oscar settings file, to accompany the status pipelines. We can then access them from `settings` instead of having to work with raw strings.

extensions/order/processing.py
```diff
@@ -0,0 +1,31 @@
...
+
+class EventHandler(processing.EventHandler):
+    """ Handles Order Processing
+
+    An override from Django Oscar for processing orders through the Fulfillment API.
+
+    """
+
+    def handle_shipping_event(self, order, event_type, lines,
+                              line_quantities, **kwargs):
```
rlucioni: Why does this method require `order` and `lines` to be provided separately? `lines` are a property of `order`, so it seems like supplying `order` should be sufficient.

extensions/order/processing.py
```diff
@@ -0,0 +1,31 @@
...
+        self.validate_shipping_event(
+            order, event_type, lines, line_quantities, **kwargs)
+
+        order = fulfillment_api.fulfill_order(order, lines)
```
rlucioni: It looks like a fulfillment attempt will occur regardless of the value provided for `event_type`; is that true? If so, it might be good to either 1) add back the check of the `event_type` you had earlier, to make the behavior more explicit, or 2) remove the `event_type` argument, if it's not necessary.

extensions/fulfillment/api.py
stephensanchez: ah, must have been a cut and paste. I'll switch to format here, but for logging exceptions the convention is to use the logger's built in string formatting, which requires %s

extensions/fulfillment/api.py
stephensanchez: It's a good question. I'm of the mind that the entire order should have its status set to "Fulfillment Error" if any part of it has an error, but not to have the API try and revoke the entirely of the order if any part fails. By design, single line items in an order could be fulfilled or revoked manually, and an atomic action across all lines would go against that contract.

extensions/fulfillment/api.py
stephensanchez: That seems like a good idea. I'll see if I can build that out.

extensions/fulfillment/api.py
rlucioni: Oh, I didn't realize that. Is there a good reason for doing that? We tend to use `format` elsewhere.

extensions/order/processing.py
stephensanchez: More to the point I was making above, this is because you can fulfill, or ship, individual lines in an order. Just as you could revoke single lines.

Imagine someone purchasing two verified seats in a single order. They may want a refund for one item, not both.

extensions/fulfillment/api.py
rlucioni: I like the sound of that.

extensions/order/processing.py
rlucioni: Got it, thanks for the context.

extensions/order/processing.py
stephensanchez: The "handle_shipping_event" function is a contract defined by Oscar that we are overriding, so I'm hesitant to change the contract. Right now, there is only one event_type, but we may have additional types in the future. It could be worth having an exception raised from here if the event type is not "Shipped", to ensure the event does not get over loaded.

rlucioni: Sure, that sounds like a good idea.

extensions/fulfillment/api.py
stephensanchez: On edx-platform it's now part of the pylint evaluation. I believe the reasoning is that the embedded format is safer

```
log.exception("Some bad stuff related to %s", my_variable) 
```

rlucioni: Just out of curiosity, what's safer about that as opposed to `log.exception("Some bad stuff related to {}".format(my_variable))`?

dianakhuang: @rlucioni I believe that it's the standard because it's cleaner and all of the logging handler stuff supports string formatting anyway.

rlucioni: Alright, makes sense. Thanks!

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,101 @@
...
+    if ORDER.COMPLETE not in order.available_statuses():
+        raise errors.IncorrectOrderStatusError(
+            "Order has a current status of [{status}] which cannot be fulfilled.".format(status=order.status)
+        )
+    modules = getattr(settings, 'FULFILLMENT_MODULE_MAPPING', {})
+
+    # Construct a dict of lines by their product type.
+    line_items = {}
+    for line in lines.all():
+        line_items.setdefault(line.product.product_class.name, []).append(line)
```
dianakhuang: I'm guessing `line.product.product_class.name` is a string. Why is the default an empty list, then?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Once all line items are organized by type, we can iterate over the Fulfillment Modules
+    # defined in our configuration and fulfill line items in the order they are designated by
+    # the configuration. Remaining line items should be marked with a fulfillment error since
+    # we have no configuration that allows them to be fulfilled.
+    for cls_path, product_types in modules.iteritems():
```
wedaly: 1) Since `FULFILLMENT_MODULE_MAPPING` is a dictionary, there's no guarantee that the rules will be processed in a particular order.  If the same product type appears in multiple rules, this could cause inconsistent behavior.

2) I think we can stop iterating through the rules once every line in the order has been fulfilled.

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,101 @@
...
+        for line in line_items[product_type]:
+            line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+            line.save()
+
+    # Check if all lines are successful, or there were errors, and set the status of the Order.
+    order_status = ORDER.COMPLETE
+    for line in lines.all():
+        if line.status != LINE.COMPLETE:
+            logger.error('There was an error while fulfilling order [%s] for user [%s]', order.id, order.user.username)
+            order_status = ORDER.FULFILLMENT_ERROR
```
wedaly: If an uncaught exception is raised earlier in this function, the line status may be saved as an error without setting the order status as an error.  Maybe put the code to set the order status in a `finally` block to be safe?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,101 @@
...
+    # the configuration. Remaining line items should be marked with a fulfillment error since
+    # we have no configuration that allows them to be fulfilled.
+    for cls_path, product_types in modules.iteritems():
+        if set(product_types).issubset(set(line_items.keys())):
+            related_lines = []
+            for product_type in product_types:
+                related_lines.extend(line_items.pop(product_type))
+            try:
+                module_path, _, name = cls_path.rpartition('.')
+                module = getattr(importlib.import_module(module_path), name)
```
wedaly: Does `importlib` cache the modules it loads?

extensions/fulfillment/api.py
```diff
@@ -0,0 +1,101 @@
...
+                    line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+                    line.save()
+
+    # Check to see if any line items in the order have not been accounted for by a FulfillmentModule
+    # Any product does not line up with a module, we have to mark a fulfillment error.
+    for product_type in line_items.keys():
+        logger.error("Product Type [%s] in order does not have an associated Fulfillment Module", product_type)
+        for line in line_items[product_type]:
+            line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+            line.save()
```
wedaly: I'm a little bit concerned about the number of database writes this will require.  Is there any way to batch these updates to the line items?

extensions/fulfillment/api.py
stephensanchez: The idea here is that we try to get the key defined by the product_class.name, if it doesn't exist, set it to an empty list.

stephensanchez:
@rlucioni @dianakhuang @wedaly Thanks for all the comments. I cleaned things up a bit, and now the module configuration is much simpler, making the code quite a bit cleaner, to boot.

extensions/fulfillment/modules.py
```diff
@@ -0,0 +1,117 @@
...
+            order (Order): The Order associated with the lines to be fulfilled
+            lines (List of Lines): Order Lines, associated with purchased products in an Order.
+
+        Returns:
+            A supported list of unmodified lines associated with "Seat" products.
+
+        """
+        supported_lines = []
+        for line in lines:
+            if line.product.product_class.name == 'Seat':
```
dianakhuang: Is there a way we could duck type this better? It's fine if there isn't. It just feels sort of odd.

stephensanchez: Heya, not sure what you had in mind, but I'm up for suggestions.  This is a little long, I agree.

rlucioni:
:+1:

dianakhuang:
:ship:

PR #12, opened 2015-02-26
Fix coverage reports for the repo

rlucioni:
Also includes superuser creation make target. 

@stephensanchez


stephensanchez:
:+1: :+1: :+1:

PR #13, opened 2015-02-26
WIP: Implement purchase endpoint

rlucioni:
This introduces a new endpoint, '/purchase'. The endpoint expects a SKU to be provided in the POST data, which is then used to populate the user's basket with the corresponding product, freeze that basket, and prepare an order using that basket. If the order total is zero (i.e., the ordered product was free), the order is then fulfilled. The endpoint returns JSON data containing order details.

There are a few additional tests I'd like to implement (one of which depends on the fulfillment API), but I figured I'd put this up now to get some eyes on it. @stephensanchez, could you take a look when you have a chance?

FYI @dianakhuang @wedaly @rock345


rlucioni:
This is no longer a WIP; I'm closing this PR in favor of #15.

PR #14, opened 2015-02-27
Implementation of the EnrollmentFulfillmentModule

stephensanchez:
Related to JIRA Ticket:
https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=109&view=detail&selectedIssue=XCOM-104

This is the implementation of the EnrollmentFulfillmentModule. I want to do a little more cleanup once it is completely wired into the Enrollment API but this branch should allow end to end testing.

FYI @dianakhuang @rlucioni


extensions/fulfillment/modules.py
```diff
@@ -111,7 +121,48 @@ def fulfill_product(self, order, lines):
...
+                "ENROLLMENT_API_HOST, ENROLLMENT_API_USERNAME, and ENROLLMENT_API_PASSWORD must be set to "
+                "use the EnrollmentFulfillmentModule"
+            )
+            for line in lines:
+                line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+
+        student = order.user.username
+        for line in lines:
+            try:
+                certificate_type= line.product.attribute_values.get(attribute__name="certificate_type").value
```
rlucioni: Nitpick: missing space between the variable and the `=`.

We should have Travis run quality checks (pep8 and pylint) as part of its test script. Right now, it only runs tests.

extensions/fulfillment/tests/test_modules.py
```diff
@@ -1 +1,151 @@
...
+        key_value.save()
+
+        # Attempt to enroll.
+        EnrollmentFulfillmentModule().fulfill_product(self.order, list(self.order.lines.all()))
+        self.assertEqual(LINE.COMPLETE, self.order.lines.all()[0].status)
+
+    @mock.patch('requests.sessions.Session.get')
+    @mock.patch('requests.cookies.RequestsCookieJar.__getitem__')
+    @mock.patch('requests.sessions.Session.post')
+    def test_enrollment_module_server_error(self, mock_get_request, mock_cookies, mock_post_request):
```
rlucioni: It doesn't look like you're using `mock_get_request` in this test.

extensions/fulfillment/tests/test_modules.py
```diff
@@ -1 +1,151 @@
...
+    @mock.patch('requests.sessions.Session.get')
+    @mock.patch('requests.cookies.RequestsCookieJar.__getitem__')
+    @mock.patch('requests.sessions.Session.post')
+    def test_enrollment_module_fulfill(self, mock_get_request, mock_cookies, mock_post_request):
+        """Happy path test to ensure we can properly fulfill enrollments. """
+        fake_enrollment_api_response = Response()
+        fake_enrollment_api_response.status_code = 200
+        mock_cookies.return_value = 'CSRF_TOKEN'
+        mock_get_request.return_value = fake_enrollment_api_response
+        mock_post_request.return_value = fake_enrollment_api_response
```
rlucioni: You might be able to clean this up a bit by moving creation of the `fake_enrollment_api_response` to the `setUp` and setting return values for your mocks in the decorators. For example:

```python
@mock.patch('requests.sessions.Session.get', mock.Mock(return_value=fake_enrollment_api_response)
@mock.patch('requests.cookies.RequestsCookieJar.__getitem__', mock.Mock(return_value='CSRF_TOKEN')
@mock.patch('requests.sessions.Session.post', mock.Mock(return_value=fake_enrollment_api_response)
```

extensions/fulfillment/modules.py
```diff
@@ -111,7 +121,48 @@ def fulfill_product(self, order, lines):
...
+                'username': username,
+                'password': password,
+                'csrfmiddlewaretoken': csrftoken,
+                'mode': certificate_type,
+                'course_details': {
+                    'course_id': course_key
+                }
+            }
+            response = client.post(url, json.dumps(data))
+            if response.status_code == 200:
```
rlucioni: Could you swap this raw status code out for one of DRF's `status` constants?

extensions/fulfillment/modules.py
```diff
@@ -111,7 +121,48 @@ def fulfill_product(self, order, lines):
...
+                continue
+
+            url = "{base_url}/api/enrollment/v1/enrollment/{user},{course_key}".format(
+                base_url=base_url, user=student, course_key=course_key
+            )
+
+            # Get the session, set the cookie.
+            client = requests.session()
+            client.get(url)
+            csrftoken = client.cookies['csrf']
```
wedaly: Are we making this GET request just to retrieve the CSRF token?

stephensanchez: Hmm, that's a good point. If I can get away with CSRF that would definitely be the right idea here.

extensions/fulfillment/modules.py
stephensanchez: I was debating it. I imagine I could. This doesn't use DRF yet, but I like those constants.

rlucioni: The purchase endpoint already requires DRF, so it wouldn't really be adding a new dependency.

extensions/fulfillment/tests/test_modules.py
stephensanchez: The get request in the function does need to be mocked out. I may need to look into this more. It works as-is but the function of the mock is not explicit.

extensions/fulfillment/tests/test_modules.py
stephensanchez: I getcha, but these fake responses are only used at a 200 response code in this test, whereas its 500 response code in the next. Since they have different functions across different tests I'd rather keep them here.

rlucioni: That's fine. What I meant was you could create the response object itself once in the setup, then set the status code on the object (200 or 500) in each test. That said, I agree with keeping things separate if you feel it makes your intentions clearer.

extensions/fulfillment/tests/test_modules.py
rlucioni: Is the request actually being made? I don't think patching the method without specifying a side effect is enough to mock out the underlying implementation. Specifying a return value should continue to allow network calls to be attempted, but the return values are then replaced with whatever you provided.

stephensanchez:
closing this and re-opening on a merge to master.

PR #15, opened 2015-02-27
Implement purchase endpoint

rlucioni:
This introduces a new endpoint, `/purchase`. The endpoint expects a SKU to be provided in POST data, which is then used to populate the user's basket with the corresponding product, freeze that basket, and prepare an order using that basket. If the order total is zero (i.e., the ordered product was free), the order is then fulfilled via the fulfillment API. The endpoint returns serialized order details.

@stephensanchez and @wedaly, could you please take a look when you have a chance?

Heads up @dianakhuang and @rock345.


extensions/purchase/tests/test_integration.py
```diff
@@ -0,0 +1,72 @@
...
+            product_class=self.product_class,
+            stockrecords=None,
+        )
+
+    def tearDown(self):
+        # Remove logger override
+        logging.disable(logging.NOTSET)
+
+    @skip('Enable once fulfillment API is available.')
+    def test_purchase_free_product(self):
```
rlucioni: This test should bring coverage close to 100%. Since it's an integration test, it requires the fulfillment API. Once the fulfillment API is merged, we can modify this test as necessary and remove the `skip` decorator.

extensions/purchase/views.py
```diff
@@ -0,0 +1,130 @@
...
+            return Response({'reason': error.message}, status=status.HTTP_400_BAD_REQUEST)
+
+        # NOTE: `get_basket` utility borrowed from the Oscar REST API.
+        basket = get_basket(request)
+        availability = basket.strategy.fetch_for_product(product).availability
+
+        # NOTE: If an exception is raised before order creation but after basket creation,
+        # an empty basket for the user will be left in the system. However, if this user
+        # attempts purchase again, the `get_basket` utility will merge all old baskets with
+        # a new one, returning a fresh basket.
```
wedaly: Is it possible that items might be included in the basket from old orders?  We avoid this in the current system by clearing the basket whenever a user starts the purchase flow.  I'm wondering if maybe we need to wrap some of this in a transaction, so that either the basket gets to a "frozen" state or it doesn't get written at all.

extensions/purchase/views.py
```diff
@@ -0,0 +1,130 @@
...
+                "number": 100021, 
+                "status": "Open", 
+                "total_excl_tax": 0.0
+            }
+        """
+        try:
+            product = get_product(request.DATA)
+        except (SkuNotFoundError, ProductNotFoundError) as error:
+            logger.error(error.message)
+            return Response({'reason': error.message}, status=status.HTTP_400_BAD_REQUEST)
```
wedaly: We've talked about adopting a convention for APIs where both a user-facing error message and a developer error message are returned.  (See [here](https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions).) I think that would be useful here.

extensions/purchase/views.py
rlucioni: Products from "old" (i.e., successfully created) orders can't be included in the newly retrieved basket at the start of a new "purchase." The `get_basket` utility [only retrieves editable baskets](https://github.com/django-oscar/django-oscar-api/blob/master/oscarapi/basket/operations.py#L65) &ndash; those with a status of `Open` or `Saved`. Baskets with a status of `Frozen` or `Submitted` are not retrieved, and since this endpoint creates new orders iff the basket in use is frozen first, items which end up being ordered can't end up in a "fresh" basket at the start of a purchase.

This assumes that the following block is atomic:
```python
basket.add_product(product)
basket.freeze()
```
We could wrap this in a transaction, but I think it's unnecessary. If there were problems with either the product or the basket, they would have been caught earlier, before attempting these operations.

extensions/purchase/views.py
rlucioni: Good idea, I've added that. The API convention docs say that user-facing error messages should be translated, so I've also taken the liberty of setting up this repo for i18n.

extensions/purchase/app.py
```diff
@@ -0,0 +1,11 @@
+from oscar.core.application import Application
+
+from purchase.urls import urlpatterns
+
+
+class PurchaseApplication(Application):
```
stephensanchez: nitpick, may want to add a docstring here.

extensions/purchase/errors.py
```diff
@@ -0,0 +1,34 @@
+"""Errors thrown by the purchase endpoint."""
+from django.utils.translation import ugettext_lazy as _
+
+
+SKU_NOT_FOUND_DEVELOPER_MESSAGE = u"No SKU provided."
```
stephensanchez: no i18n on this message.

stephensanchez:
This is looking pretty clean, Renzo! I think you and I both need to add a few more log messages in to capture info-level but otherwise I think we're going to be good for wiring a few things together on Monday. :+1:

extensions/purchase/errors.py
rlucioni: Thanks! The [API convention docs](https://openedx.atlassian.net/wiki/display/AC/edX+REST+API+Conventions) state that developer messages (like this one) shouldn't be i18n'd; only those messages that might surface in a UI should be i18n'd.

extensions/purchase/app.py
rlucioni: All set.

rlucioni:
Sounds good, thanks for the review @stephensanchez. Maybe we can open a separate PR on Monday to drop in the info-level logging once this endpoint and the fulfillment API are in place, and we've been able to talk about what points in the process we want to be logging at. I'm also planning to open another separate PR to introduce rate-limiting for this endpoint; maybe we could talk about that on Monday as well?

wedaly:
:+1: Take it to merge-town!

rlucioni:
Next stop, Mergetown.

PR #16, opened 2015-03-01
Accelerate test runs by disabling migrations

rlucioni:
Disabling migrations lowers test runtime from 21 seconds to 5 seconds. @stephensanchez and/or @dianakhuang, what do you think?


stephensanchez:
Smart idea, any chance you could tweak it to allow migrations to run if you want them to? i.e. make test_python --run-migrations, then don't set the environment variable?

rlucioni:
From offline discussion: since test databases are blown away and recreated on a per-test basis, there shouldn't be a need to run tests with migrations. @stephensanchez, is it okay with you if I go ahead and merge this?

stephensanchez:
:+1:

PR #17, opened 2015-03-02
Implementation of the EnrollmentFulfillmentModule

stephensanchez:
Related to JIRA Ticket:
https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=109&view=detail&selectedIssue=XCOM-104

This is the implementation of the EnrollmentFulfillmentModule. I want to do a little more cleanup once it is completely wired into the Enrollment API but this branch should allow end to end testing.

I will circle back with @mattdrayer to get some ideas about CSRF and how we'll want to handle the endpoint communication to the Enrollment API.


stephensanchez:
@dianakhuang @rlucioni @wedaly FYI I re-created this PR against master. Just tested it with @dianakhuang 's changes and it is working.

extensions/fulfillment/tests/test_modules.py
```diff
@@ -1 +1,145 @@
...
+
+import ddt
+import mock
+from nose.tools import raises
+from django.test import TestCase, override_settings
+from django.contrib.auth.models import User
+from requests import Response
+from rest_framework import status
+from extensions.fulfillment.modules import FulfillmentModule, EnrollmentFulfillmentModule
+from extensions.fulfillment.status import ORDER, LINE
```
rlucioni: Nitpick: could you separate these local `extensions` imports from other third-party imports with a newline? This is the import grouping [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) recommends.

extensions/fulfillment/modules.py
```diff
@@ -4,6 +4,17 @@
...
 
 """
+import json
+import logging
+import requests
+
+from django.conf import settings
+from rest_framework import status
+
+from extensions.fulfillment.status import LINE
```
rlucioni: Same comment here on import grouping; could you group `from oscar.apps.catalogue.models import ProductAttributeValue` with the other third-party imports here and give the import from `extensions` its own group?

rlucioni:
Nice! Sweet coverage, too. Just a few comments, otherwise :+1: .

extensions/fulfillment/modules.py
```diff
@@ -111,7 +122,43 @@ def fulfill_product(self, order, lines):
...
+        for line in lines:
+            try:
+                certificate_type = line.product.attribute_values.get(attribute__name="certificate_type").value
+                course_key = line.product.attribute_values.get(attribute__name="course_key").value
+            except ProductAttributeValue.DoesNotExist:
+                logger.error("Supported Seat Product does not have required attributes, [certificate_type, course_key]")
+                line.set_status(LINE.FULFILLMENT_CONFIGURATION_ERROR)
+                continue
+
+            url = "{base_url}api/enrollment/v1/enrollment".format(base_url=base_url)
```
dianakhuang: Does it make sense to hardcode this url? I can see the argument that if this changes everything is going to have to change, but I feel odd having all this logic exist in `fulfill_product` instead of a separate method/function.

extensions/fulfillment/modules.py
```diff
@@ -111,7 +122,43 @@ def fulfill_product(self, order, lines):
             The original set of lines, with new statuses set based on the success or failure of fulfillment.
 
         """
-        pass
-
-    def revoke_product(self, order, lines):
-        pass
\ No newline at end of file
+        base_url = getattr(settings, 'ENROLLMENT_API_HOST', '')
+        api_key = getattr(settings, 'EDX_API_KEY', '')
```
rlucioni: I might make this more explicit by using `None` as the default value, instead of the empty string.

extensions/fulfillment/modules.py
rlucioni: It seems appropriate to at least put the path in settings, so that it can be updated without having to touch the code.

stephensanchez: Makes sense to me. I'm going to simply create ENROLLMENT_API_URL and require the full path and host specified. Simplifies things a bit.

stephensanchez:
@rlucioni @dianakhuang review comments addressed!

rlucioni:
Looks good @stephensanchez, thanks for addressing those comments. :+1:

extensions/settings/_oscar.py
```diff
@@ -65,6 +65,12 @@
     'extensions.fulfillment.modules.EnrollmentFulfillmentModule',
 ]
 
+# EnrollmentFulfillmentModule Settings
+ENROLLMENT_API_URL = "http://localhost:8000/api/enrollment/v1/enrollment"
+
+# This key is used for access to the Enrollment API without a designated user.
+EDX_API_KEY = None
```
dianakhuang: How is this meant to be overridden on prod/other environments?

stephensanchez: I'm going to catch up with @feanil but I imagine we're going to need an auth.json and env.json file similar to platform, then include that here. I could set that up within this PR but I imagine the files themselves will need to exist in the ecomstack

dianakhuang: Okay. All I'm familiar with is how `edx-platform` handles things like this, where you have to add an entry to `aws.py` that specifically reads the data from the JSON files. If we don't have a similar settings file yet, that's fine, but I just wanted to make sure that we weren't forgetting to add it to our equivalent of `aws.py`.

stephensanchez: Actually I think it'd make sense for this to be a separate PR in the edx-ecommerce repo, that may override these settings. I'll create something for this. @dianakhuang

feanil: Please don't add an aws.py. Let's talk before any of those settings things happen.  The new pattern is different from what edx-platform does and is much simpler.  I can walk you guys through it tomorrow.

rlucioni: @feanil I'd like to be part of that conversation, too, so I can understand how this works.

PR #18, opened 2015-03-02
Add throttling to the purchase endpoint

rlucioni:
I've used DRF to add throttling to the purchase endpoint. At the moment, I've chosen to limit requests to the purchase endpoint to 15 per minute, per user. I've also included a unit test for verifying that throttling is working as expected.

The second commit in this PR cleans up this repo's README.

@stephensanchez and/or @dianakhuang, could you review this when you have a moment?


README.rst
```diff
@@ -1,39 +1,66 @@
...
+Documentation |ReadtheDocs|_ 
+------------------------------
+.. |ReadtheDocs| image:: https://readthedocs.org/projects/edx-django-oscar-extensions/badge/?version=latest
+.. _ReadtheDocs: http://edx-django-oscar-extensions.readthedocs.org/en/latest/
 
 License
 -------
 
-The code in this repository is licensed under LICENSE_TYPE unless
+The code in this repository is licensed under the AGPL unless
```
stephensanchez: Thanks for the README cleanup.

extensions/purchase/views.py
```diff
@@ -22,8 +23,15 @@
 EventHandler = get_class('order.processing', 'EventHandler')
 
 
+class PurchaseThrottle(UserRateThrottle):
+    """Limit the number of requests users can make to the purchase endpoint."""
+    rate = '15/minute'
```
stephensanchez: This may be too conservative. Track selection for verified certificates will eventually make two purchases per course, if you first make a free honor cert purchase, then a verified cert purchase. That means if someone clicks through and purchases an X-Series or a bunch of courses, they may get denied on the 7th course. 

May be worth increasing this a bit, just to avoid someone being explicitly malicious. 40/minute, perhaps?

extensions/purchase/tests/test_views.py
```diff
@@ -131,6 +132,16 @@ def test_no_product_for_sku(self):
             )
         )
 
+    def test_throttling(self):
```
stephensanchez: Cool way to test the throttling. I'd say keep it in now that you've created the test, but in the future I'd say it's not necessary. This test is exercising the DRF more than our code, in my opinion.

extensions/purchase/tests/test_integration.py
```diff
@@ -21,8 +21,6 @@
 
 Order = get_model('order', 'Order')
 ShippingEventType = get_model('order', 'ShippingEventType')
-Free = get_class('shipping.methods', 'Free')
-OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
```
dianakhuang: Just out of curiosity. Why were these here in the first place?

extensions/purchase/views.py
rlucioni: Sounds good. Looking at the minute-by-minute enrollment event totals in GA, 40/minute seems a bit too generous to me, but it'll definitely work as a ceiling to catch those hammering the endpoint.

extensions/purchase/tests/test_integration.py
rlucioni: Left over from an old test I later refactored.

extensions/purchase/tests/test_views.py
rlucioni: Thanks, and good point. I thought it'd be good to have this in place in case someone accidentally broke the throttling while making changes.

rlucioni:
@stephensanchez I think we're all set here. The rate limit is now at 40/minute, and is also configurable from settings (`_oscar.py`). I refactored the test a little to avoid the need to make `2n+1` requests, where `n` is the rate limit; it now only makes `n+1` requests.

stephensanchez:
:+1:

PR #20, opened 2015-03-03
Updating info level logging for Fulfillment API

stephensanchez:
@rlucioni 

This is adding a number of fulfillment API info log level messaging, and revisiting some of the old error messages. Removed the username from one of the messages as well.


extensions/fulfillment/modules.py
```diff
@@ -156,7 +158,10 @@ def fulfill_product(self, order, lines):
 
             response = requests.post(enrollment_api_url, data=json.dumps(data), headers=headers)
             if response.status_code == status.HTTP_200_OK:
+                logger.info("Success fulfilling line [%d] of order [%d].", line.id, order.id)
                 line.set_status(LINE.COMPLETE)
             else:
+                logger.error("Unable to fulfill line [%d] of order [%d] due to server side error.", line.id, order.id)
                 line.set_status(LINE.FULFILLMENT_SERVER_ERROR)
+        logger.info("Finished fulfilling 'Seat' product types for order id [%d]", order.id)
```
rlucioni: These messages sometimes refer to `order [%d]` and sometimes to `order id [%d]`. I think it's pretty clear that the value in the brackets is an ID; the log messages from my end all refer to `item [number]`. For the sake of consistency, could you modify the messages so that they read the same way?

extensions/fulfillment/api.py
```diff
@@ -70,11 +71,12 @@ def fulfill_order(order, lines):
         for line in lines.all():
             if line.status != LINE.COMPLETE:
                 logger.error(
-                    'There was an error while fulfilling order [%s] for user [%s]', order.id, order.user.username
+                    'There was an error while fulfilling order [%d] for user id [%d]', order.id, order.user.id
```
rlucioni: Is it necessary to log a user ID here? Having the order number logged should be enough to tie it back to a user.

rlucioni:
I think we should be logging order numbers instead of order IDs (primary keys). You and I are both logging IDs instead of order numbers. What do you think? If you agree, I can make a quick change to the log messages in the purchase endpoint.

stephensanchez:
Yes on all accounts. The order # makes much more sense, and I can drop the
user ID.

rlucioni:
Cool. I've created #22 to bring in the changes to log messages in the purchase endpoint.

stephensanchez:
@rlucioni updated accordingly.

rlucioni:
:+1:

PR #21, opened 2015-03-03
Clean up import paths

rlucioni:
Using fully qualified import paths allows us to use extensions from the ecommerce service. @stephensanchez, could you take a quick look?


stephensanchez:
:+1:

PR #22, opened 2015-03-03
Log order numbers instead of order IDs

rlucioni:
@stephensanchez, as discussed. It doesn't look like baskets have a number associated with them like orders do, so I think we have to keep using IDs for them.


stephensanchez:
:+1:

stephensanchez:
Oh one thing, the order.number is a string, not a digit.

rlucioni:
Oh, good call. I've updated the string formatting and the order serializer, too.

PR #23, opened 2015-03-03
Change purchase API to update order line status.

stephensanchez:
A small change to ensure that the order line status is updated when the order status is updated. By calling set_status() on the order, it will invoke the status cascade defined for the order / line relationship.

@rlucioni


extensions/purchase/views.py
```diff
@@ -162,6 +162,11 @@ def _prepare_order(self, basket, product, sku):
             basket.id,
         )
 
+        if order.total_excl_tax == self.FREE:
+            order.set_status(ORDER.BEING_PROCESSED)
+            order.set_status(ORDER.PAID)
+            logger.info(u"Marking order number [%s] as [%s]", order.number, ORDER.PAID)
```
rlucioni: Could you change this message to `Marking order [%s] as...`? It's the same consistency thing I mentioned earlier; we don't need to say "number."

rlucioni:
Just a nitpick, otherwise :+1: .

PR #26, opened 2015-03-13
Protect Oscar front-end

rlucioni:
Makes Oscar URLs available only to users with staff permissions, and makes the management dashboard available at the root. Accompanied by [#23](https://github.com/edx/ecommerce/pull/23) in the edx/ecommerce repo.

After reading through Oscar's Google Groups archives and much experimenting, it's become evident that removing the URLs for Oscar's front-end isn't the most elegant way to solve the problem at hand. It would require revising templates and parts of the Oscar core which assume that these URLs exist. However, a clean way to, in effect, disable these URLs is to only make them available to users with staff permissions, the same protection used to guard the management dashboard from public access.

The changes in this PR render the extensions project unusable from a browser while running outside of the ecommerce service, since the default Oscar login page is made available only to users with staff permissions :expressionless:. However, since the ecommerce service relies on the LMS login page for authentication and not on the Oscar login page, this minimally invasive approach allows us to protect Oscar's front-end without sacrificing any internal functionality.

@clintonb and/or @dianakhuang, could you please review this when you're able?


clintonb:
Magic! :+1:

Makefile
```diff
@@ -27,9 +26,6 @@ dev_requirements:
 test_requirements:
 	pip install -qr requirements/test.txt --exists-action w
 
-super:
-	python manage.py createsuperuser
-
```
dianakhuang: What's the reason for removing this? Just assuming that people will use their LMS credentials that have staff permissions?

rlucioni: Pretty much. As I told @clintonb in an earlier PR, the ecommerce service uses the LMS to authenticate and create users with their LMS permissions preserved. It doesn't seem to make much sense to encourage superuser creation by supplying a make target for it if you won't be able to log in as the superuser you created. If you need a superuser for the ecommerce service, you'll just log in as a superuser from the LMS.

dianakhuang:
:rocket:

PR #28, opened 2015-03-18
Update product type expectations

stephensanchez:
Django Oscar products will have a type "None" if they are child products. All of our products are children, so build a utility function to check the parent product type as needed to map to fulfillment modules.

Review when able!
@dianakhuang @clintonb @rlucioni


rlucioni:
Thanks for the quick fix! :+1:

clintonb:
:+1:

PR #29, opened 2015-03-18
Increment version

rlucioni:
Required before building a new AMI. I will tag a new release after merging this. @clintonb or @stephensanchez.


clintonb:
:+1:

PR #30, opened 2015-03-18
Logging additional details when fulfillment fails

clintonb:
@rlucioni @dianakhuang @stephensanchez


rlucioni:
:+1:

Don't forget to update the ecommerce service to install this new version.

clintonb:
Will do. Waiting for the build to finish before merging.

extensions/fulfillment/modules.py
```diff
@@ -182,8 +182,15 @@ def fulfill_product(self, order, lines):
                     logger.info("Success fulfilling line [%d] of order [%s].", line.id, order.number)
                     line.set_status(LINE.COMPLETE)
                 else:
+                    try:
+                        data = response.json()
+                        reason = data.get('message')
+                    except Exception:   # pylint: disable=broad-except
+                        reason = '(No detail provided.)'
```
dianakhuang: Is there a particular reason why this message is in parens?

dianakhuang:
Looks like you've angered Coveralls, but otherwise, :rocket:

rlucioni:
I'm forced to withdraw my thumb until Coveralls is happy.

extensions/fulfillment/modules.py
clintonb: To indicate that the message is coming from our code, not the API.

clintonb:
Updated.

clintonb:
Updated.

stephensanchez:
:+1:

PR #31, opened 2015-03-18
Payment processor transaction parameter generation

rlucioni:
Includes a payment processor class for generating CyberSource transaction parameters. The order endpoint now returns CyberSource transaction parameters for newly created orders. Whether to use them is up to the LMS. This meets the requirements of [XCOM-114](https://openedx.atlassian.net/browse/XCOM-114).

@clintonb and @stephensanchez, could you please review this when you're able?


stephensanchez:
This looks really clean, Renzo. If we have 30 minutes today I'd love a walk through of this, though, because I'm not familiar with CyberSource. If able, maybe we can get @wedaly to review this quickly as well?

rlucioni:
@stephensanchez good idea, I should have thought to tag @wedaly. I'd be happy to walk you through it.

extensions/payment/helpers.py
```diff
@@ -0,0 +1,45 @@
...
+
+    Arguments:
+        path (string): Fully-qualified path to a payment processor class.
+
+    Returns:
+        class: The payment processor class at the specified path.
+
+    Raises:
+        ImportError: If no module with the parsed module path exists.
+        AttributeError: If the module located at the parsed module path does not contain a
```
wedaly: Nitpick: small typo here "a \n a"

extensions/settings/_oscar.py
```diff
@@ -132,3 +132,12 @@
 ########## RATE LIMITING
 ORDERS_ENDPOINT_RATE_LIMIT = '40/minute'
 ########## END RATE LIMITING
+
+
+########## PAYMENT PROCESSING
+PAYMENT_PROCESSOR_CLASS = 'extensions.payment.processors.Cybersource'
```
wedaly: This isn't critical yet, but very soon the E-Commerce service is going to need to support multiple payment gateways (CyberSource and Paypal).  We might as well begin building the system with that assumption in mind.

Instead of defining a single payment processor class, I'd recommend generalizing this to a dictionary.  The keys of the dictionary can be processor names ("cybersource", "paypal", etc).  We'll need these names down the road when we begin defining receipt/cancellation/fulfillment end-points that understand the responses sent by each processor.

extensions/api/views.py
```diff
@@ -201,3 +204,13 @@ def _fulfill_order(self, order, shipping_event):
 
         fulfilled_order = EventHandler().handle_shipping_event(order, shipping_event, order_lines, [line_quantity])
         return fulfilled_order
+
+    def _assemble_order_data(self, order):
+        """Assemble a dictionary of metadata for the provided order."""
+        order_data = serializers.OrderSerializer(order).data
+
+        payment_processor = get_processor_class(getattr(settings, 'PAYMENT_PROCESSOR_CLASS'))
```
wedaly: This is fine for now, but eventually we're going to need to extend the end-point so that it accepts an optional payment processor implementation (so we can have a button for PayPal in addition to the usual payment method).  Maybe have a setting to specify the default processor implementation?

extensions/api/views.py
```diff
@@ -127,9 +129,9 @@ def post(self, request):
             shipping_event = data.get_shipping_event(self.SHIPPING_EVENT_NAME)
             order = self._fulfill_order(order, shipping_event)
 
-        order_serializer = serializers.OrderSerializer(order)
+        order_data = self._assemble_order_data(order)
```
wedaly: I think we're going to want to (a) save which payment processor we used for the order somewhere in the database, and (b) include info-level logging about which payment processor we used.  When we support multiple processors, we're going to need to know which order is paid for in which system.

wedaly:
I'm realizing as I'm writing that supporting multiple payment processors is probably enough work to justify a separate story.  I'm okay with merging this PR now if you can create a JIRA ticket to track the work for extending this to support multiple gateways.

:+1:

extensions/payment/errors.py
```diff
@@ -0,0 +1,14 @@
+"""Exceptions and error messages used by payment processors."""
+class ProcessorError(Exception):
```
clintonb: Not sure what other processors may come in the future; but, we can avoid future confusion by calling this `PaymentProcessorError`.

extensions/payment/processors.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_transaction_parameters(order, receipt_page_url=None, cancel_page_url=None, merchant_defined_data=None):
+        """Generate a dictionary of transaction parameters to be sent to a payment processor."""
+        raise NotImplementedError("Transaction parameters method not implemented.")
+
+
+class Cybersource(BasePaymentProcessor):
+    """CyberSource Secure Acceptance Web/Mobile (February 2015)
+
+    For reference, see http://goo.gl/KNXsK.
```
clintonb: Link directly to http://apps.cybersource.com/library/documentation/dev_guides/Secure_Acceptance_WM/Secure_Acceptance_WM.pdf.

extensions/payment/processors.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+    NAME = CS.NAME
+
+    def get_transaction_parameters(self, order, receipt_page_url=None, cancel_page_url=None, merchant_defined_data=None):
+        """Generate a dictionary of signed parameters CyberSource requires to complete a transaction.
```
clintonb: Move this docstring to the base class? Willy pep8/pylint complain?

extensions/payment/processors.py
```diff
@@ -0,0 +1,181 @@
...
+            OrderedDict: CyberSource-specific parameters required to complete a transaction. An OrderedDict is
+                used to facilitate testing.
+
+        Raises:
+            AttributeError: If settings.CYBERSOURCE_ACCESS_KEY, settings.CYBERSOURCE_PROFILE_ID,
+                or settings.LANGUAGE_CODE is not set.
+            ExcessiveMerchantDefinedData: If the provided merchant-defined data exceeds CyberSource's
+                optional field limit.
+        """
+        parameters = OrderedDict()
```
clintonb: How does an `OrderDict` facilitate testing? I'm hesitant to change implementation to facilitate testing.

extensions/payment/processors.py
```diff
@@ -0,0 +1,181 @@
...
+    def _add_signature(self, parameters):
+        """Sign the contents of the provided transaction parameters dictionary.
+
+        This allows CyberSource to verify that the transaction parameters have not been tampered with
+        during transit. The parameters dictionary should contain a key 'signed_field_names' which CyberSource
+        uses to validate the signature. The message to be signed must contain parameter keys and values ordered
+        in the same way they appear in 'signed_field_names'.
+
+        Arguments:
+            parameters (OrderedDict): A dictionary of transaction parameters.
```
clintonb: This argument doesn't *need* to be an `OrderDict`.

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+class PaymentProcessorTests(TestCase):
+    """Base test class for payment processor classes."""
+    ORDER_NUMBER = '001'
+
+    def setUp(self):
+        # Override all loggers, suppressing logging calls of severity CRITICAL and below
+        logging.disable(logging.CRITICAL)
+
+        user = User.objects.create_user(
+            username='Gus', email='gustavo@lospolloshermanos.com', password='the-chicken-man'
```
clintonb: Nice!

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+
+class PaymentProcessorTests(TestCase):
+    """Base test class for payment processor classes."""
+    ORDER_NUMBER = '001'
+
+    def setUp(self):
+        # Override all loggers, suppressing logging calls of severity CRITICAL and below
+        logging.disable(logging.CRITICAL)
+
+        user = User.objects.create_user(
```
clintonb: Minor: Use django-dynamic-fixture (since we have it).

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+            title=u'𝑭𝒓𝒊𝒆𝒅 𝑪𝒉𝒊𝒄𝒌𝒆𝒏',
+            product_class=product_class,
+            stockrecords=None,
+        )
+        pollos_hermanos = factories.ProductFactory(
+            structure='child',
+            parent=fried_chicken,
+            title=u'𝕃𝕠𝕤 ℙ𝕠𝕝𝕝𝕠𝕤 ℍ𝕖𝕣𝕞𝕒𝕟𝕠𝕤',
+            stockrecords__partner_sku=u'ṠÖṀЁṪḦЇṄĠ⸚ḊЁḶЇĊЇÖÜṠ',
+            stockrecords__price_excl_tax=D('9.99'),
```
clintonb: Seems pretty expensive. Is this for a whole chicken?

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+User = get_user_model()
+
+
+class PaymentProcessorTests(TestCase):
+    """Base test class for payment processor classes."""
+    ORDER_NUMBER = '001'
+
+    def setUp(self):
+        # Override all loggers, suppressing logging calls of severity CRITICAL and below
+        logging.disable(logging.CRITICAL)
```
clintonb: Why no logging? Might be useful for debugging test failures.

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+    @raises(NotImplementedError)
+    def test_base_processor_unusable(self):
+        """Test that use of the base payment processor class fails due to NotImplementedError."""
+        BasePaymentProcessor().get_transaction_parameters(self.order)
+
+
+class CybersourceTests(PaymentProcessorTests):
+    """Tests of the CyberSource processor class."""
+    UUID_HEX = 'madrigal'
+    PI_DAY = datetime.datetime(2015, 3, 14, 21, 26, 53)
```
clintonb: `2015, 3, 14, 9, 26, 53`

extensions/payment/tests/test_processors.py
```diff
@@ -0,0 +1,172 @@
...
+        expected_parameters[CS.FIELD_NAMES.SIGNED_FIELD_NAMES] = CS.SEPARATOR.join(expected_parameters.keys())
+
+        # Generate a comma-separated list of keys and values to be signed. CyberSource refers to this
+        # as a 'Version 1' signature in their documentation.
+        message = CS.SEPARATOR.join([CS.MESSAGE_SUBSTRUCTURE.format(key=key, value=expected_parameters.get(key)) for key in expected_parameters.keys()])
+        secret = getattr(settings, 'CYBERSOURCE_SECRET_KEY')
+
+        expected_parameters[CS.FIELD_NAMES.SIGNATURE] = sign(message, secret)
+
+        self.assertEqual(returned_parameters, expected_parameters)
```
clintonb: I'm missing the need for an `OrderDict`.

extensions/payment/processors.py
wedaly: I believe the keys are joined into a comma separated list, which is used to generate the digital signature.  If the order of the keys is non-deterministic, then the signature might change, which could cause test failures.

I don't think this is only a testing concern though.  There's value in ensuring that the signatures we generate are 100% deterministic and reproducible :)

rlucioni: @wedaly that's correct; @clintonb, I would have responded with the above.

extensions/payment/helpers.py
rlucioni: Fixed.

rlucioni:
@wedaly I've created [XCOM-149](https://openedx.atlassian.net/browse/XCOM-149) to track the work for supporting multiple gateways.

extensions/payment/errors.py
rlucioni: Updated.

extensions/payment/processors.py
rlucioni: Updated. What's wrong with the shortened URL?

clintonb: There is nothing inherently wrong with it; but, unless we expect the URL to change frequently, there is no real need to have a proxy/shortened URL.

extensions/payment/processors.py
rlucioni: This docstring includes details specific to CyberSource. I'll expand the base class' docstring to explain the purpose of the expected arguments. That should keep pep8/pylint happy, too. (Speaking of which, we need to tell Travis to check pep8/pylint.)

clintonb: RE: Travis. We have a make target, validate, in Insights. It calls test_python and quality.

extensions/payment/processors.py
rlucioni: Good catch. Updated.

extensions/payment/tests/test_processors.py
rlucioni: :chicken:

extensions/payment/tests/test_processors.py
rlucioni: But then I can't make users with clever names and emails, like Gus!

extensions/payment/tests/test_processors.py
wedaly: pollos

extensions/payment/tests/test_processors.py
rlucioni: Based on the way I've configured the products, it looks like that's the price to buy the restaurant! Maybe that's how much you need to pay to open a franchise. Not such a bad deal, if you ask me.

extensions/payment/tests/test_processors.py
rlucioni: It's noisy. I figure breakpoints should be sufficient to debug test failures. If necessary, re-enabling logging while tests are running is easy.

extensions/payment/tests/test_processors.py
rlucioni: Technically, `2015, 3, 14, 21, 26, 53` is also correct: https://docs.python.org/2/library/datetime.html#datetime.datetime.hour.

I'll let you have this one, though.

clintonb: Boo! 9AM!

extensions/api/views.py
clintonb: I second this need. We should do this now to avoid data migrations in the future.

clintonb:
My concerns regarding `OrderDict` have been addressed. Thanks. Overall, everything looks good; but, I would like to see the payment processor recorded in the database now, to avoid data migrations in the future. Data migrations aren't bad. I'm just hesitant to touch production data, especially commerce-related data.

extensions/api/views.py
rlucioni: All set here.

rlucioni:
@wedaly and @clintonb, as suggested, I've added a `payment_processor` field to the order model, and extended the info-level log message emitted when orders are created to include information about the payment processor to be used.

I've also bumped the version of the `requests` module to 2.6.0, upgraded to fix a recently disclosed vulnerability.

extensions/payment/processors.py
rlucioni: Thanks for the tip.

rlucioni:
I've added a second commit to this PR which cleans up all pep8 and pylint violations in this project. I've also modified the Travis configuration to use a new `make` target which will run our unit tests in addition to running pep8 and pylint.

clintonb:
:+1:

PR #32, opened 2015-03-20
Upgrade to Django 1.7.7, DRF-JWT 1.4.0, PyJWT 1.0.0

rlucioni:
I'll include these upgrades as part of the 0.0.5 release of extensions.

@clintonb or @stephensanchez


stephensanchez:
:+1:

PR #33, opened 2015-03-24
Added support for retrying fulfillment

clintonb:
@rlucioni @stephensanchez


extensions/_version.py
```diff
@@ -1 +1 @@
-__version__ = '0.0.5'  # pragma: no cover
+__version__ = '0.0.6'  # pragma: no cover
```
stephensanchez: We should have a better process for bumping versions. I'd prefer to bump the extensions versions when we tag a release, on a weekly deployment basis. This is fine for this PR but just wanted to mention it.

extensions/static/oscar/js/order_list.js
```diff
@@ -0,0 +1,23 @@
...
+// This must be tested manually.
+
+$(document).ready(function () {
+    $('[data-action=retry-fulfillment]').click(function (e) {
+        var $btn = $(e.target),
+            order_number = $btn.data('order-number');
+
+        // Disable button
+        e.preventDefault();
+        $btn.addClass('disabled');
```
stephensanchez: Don't you need to enable the button again on success or failure?

extensions/templates/oscar/dashboard/orders/order_list.html
```diff
@@ -0,0 +1,197 @@
+{% extends 'dashboard/layout.html' %}
```
stephensanchez: In order to change templates, we have to replace the entire template with an over ride? Just making sure I understand it, and I think I prefer that anyhow.

stephensanchez:
One or two comments, only one thing needs to be addressed in the javascript I believe, otherwise this looks great! :+1:

extensions/templates/oscar/dashboard/orders/order_list.html
clintonb: Correct. I also added a setting, `OSCAR_EXTENSIONS_STATIC_DIR`, so that Django knows to first look for extension overrides.

extensions/static/oscar/js/order_list.js
clintonb: This is a rough first iteration. The template doesn't currently have any messaging capability to inform the user of errors. I may add that, if it isn't too difficult.

clintonb:
@rlucioni @stephensanchez @dianakhuang I added a Selenium test for the front-end. This was simpler than adding a proper JS testing framework. We will eventually need that framework; but, I'd rather not add that overhead to test this relatively simple feature. Please take one more look.

extensions/static/oscar/js/order_list.js
clintonb: I'm going to stick wit the current implementation. We can revisit the UI/UX at a later time if (when?) the lack of proper messaging around errors becomes an issue.

clintonb: Scratch that. The page now displays alerts upon success and failure. The button is removed if fulfillment succeeds, and reactivated if fulfillment fails.

extensions/api/urls.py
```diff
@@ -6,4 +6,5 @@
 urlpatterns = patterns(
     '',
     url(r'orders/$', views.OrdersView.as_view(), name='orders'),
+    url(r'orders/(?P<number>[-\w]+)/fulfill/$', views.FulfillOrderView.as_view(), name='fulfill_order'),
```
rlucioni: @dianakhuang this seems relevant to the story you're working on. This does seems like the most appropriate place for the fulfillment endpoint; it should be part of the API.

extensions/api/views.py
```diff
@@ -20,17 +22,34 @@
...
+
+class FulfillmentMixin(object):
+    SHIPPING_EVENT_NAME = 'Shipped'
+
+    def _fulfill_order(self, order):
+        """Attempt fulfillment for orders consisting of a free product."""
+        # We use `all` to get a QuerySet object to pass to the EventHandler, but the order
+        # only contains a single line item.
+        order_lines = order.lines.all()
+        line_quantities = [1] * order_lines.count()
```
rlucioni: Nice touch.

extensions/api/views.py
```diff
@@ -20,17 +22,34 @@
...
 EventHandler = get_class('order.processing', 'EventHandler')
+Order = get_model('order', 'Order')
 OrderTotalCalculator = get_class('checkout.calculators', 'OrderTotalCalculator')
-OrderCreator, OrderNumberGenerator = get_classes('order.utils', ['OrderCreator', 'OrderNumberGenerator'])  # pylint: disable=unbalanced-tuple-unpacking
+
+# pylint: disable=unbalanced-tuple-unpacking
+OrderCreator, OrderNumberGenerator = get_classes('order.utils', ['OrderCreator', 'OrderNumberGenerator'])
+
+
+class FulfillmentMixin(object):
```
rlucioni: We're doing a good job of keeping our code distributed across small, digestible files. I don't want to overdo it, but what do you think of moving this to a mixins module (`extensions.api.mixins`)?

extensions/api/urls.py
dianakhuang: I think it's wrong to think of my endpoint as a fulfillment endpoint, but I don't think this discussion really belongs in this code review.

rlucioni: Fair point, we can discuss it elsewhere.

extensions/api/views.py
```diff
@@ -20,17 +22,34 @@
...
+    SHIPPING_EVENT_NAME = 'Shipped'
+
+    def _fulfill_order(self, order):
+        """Attempt fulfillment for orders consisting of a free product."""
+        # We use `all` to get a QuerySet object to pass to the EventHandler, but the order
+        # only contains a single line item.
```
dianakhuang: Do we want to make this assumption that there's only a single item in the order? I'm just thinking forward to things like the future of bringing whitelabel products into this, which currently can have more than one item in an order.

extensions/order/processing.py
```diff
@@ -19,12 +21,34 @@ class EventHandler(processing.EventHandler):
...
 
         order = fulfillment_api.fulfill_order(order, lines)
 
-        self.create_shipping_event(
-            order, event_type, lines, line_quantities, **kwargs)
+        self.create_shipping_event(order, event_type, lines, line_quantities, **kwargs)
 
         return order
+
+    def create_shipping_event(self, order, event_type, lines, line_quantities, **kwargs):
```
dianakhuang: Docstring? An explanation for why we're overriding this?

extensions/api/views.py
clintonb: The file is still pretty small. Let's move the mixin out the next time we add an endpoint.

extensions/api/views.py
clintonb: This assumption actually caused me problems when testing. This comment is no longer relevant, and I'll remove it. My change to `line_quantities` removes this assumption, but does assume that the user is only purchasing one of everything in the basket.

```python
# Old: One item
line_quantities = 1

# New: One of every item
line_quantities = [1] * order_lines.count()
```

dianakhuang: I'm a little concerned about that assumption going forward, too. One of the features that they use for white label is the ability to bulk purchase seats in a course. For now, this is fine, but I don't know what's wrong with using the order's existing quantities? Do they just not exist?

extensions/order/processing.py
clintonb: Added.

extensions/api/views.py
clintonb: This is a DUH moment. Yes, the quantities exist on the line items. We are now using them.

extensions/api/views.py
```diff
@@ -20,17 +22,32 @@
...
+OrderCreator, OrderNumberGenerator = get_classes('order.utils', ['OrderCreator', 'OrderNumberGenerator'])
+
+
+class FulfillmentMixin(object):
+    SHIPPING_EVENT_NAME = 'Shipped'
+
+    def _fulfill_order(self, order):
+        """Attempt fulfillment for orders consisting of a free product."""
+        order_lines = order.lines.all()
+        line_quantities = [line.quantity for line in order_lines]
```
clintonb: @dianakhuang This is the fix for quantities. No more assumptions!

dianakhuang:
I think that was all my comments. :ship:

PR #34, opened 2015-03-24
Enable ATOMIC_REQUESTS for local development

rlucioni:
Wraps views in a transaction to more closely mirror production config, and removes what would be a redundant `atomic()` context manager. Use of SQLite in conjunction with `ATOMIC_REQUESTS` prevents us from disabling `AUTOCOMMIT` locally.

@stephensanchez or @clintonb, could you please take a look at this? @wedaly, tagging you since we discussed use of the `atomic()` context manager earlier. Since we're using `ATOMIC_REQUESTS` now, the entire view is an atomic operation.

@feanil, FYI.


clintonb:
:+1:

stephensanchez:
:+1:

rlucioni:
After further consideration, we've decided to leave `ATOMIC_REQUESTS` set to its default value (`False`). If view-level atomicity is necessary, we can explicitly decorate views with `@atomic`, which is preferable to assuming that all views operate within atomic blocks. To be clear, `AUTOCOMMIT` will still be disabled in production.

The changes in this PR are no longer necessary, so I'll go ahead and close it.

rlucioni:
Reopening and merging now that we understand the need for atomic requests.

rlucioni:
Pylint is reporting an `unused-import` false positive, but only when run by Travis. I'm going to close this PR again and open a new one to see if that fixes the issue.

PR #35, opened 2015-03-25
Add backend acceptance of the response from Cybersource.

dianakhuang:
Check and verify the response from CyberSource.
- Does not record anything to the DB.
- Does not trigger fulfillment

@stephensanchez @clintonb


extensions/order/models.py
```diff
@@ -6,6 +6,32 @@
...
+        Returns:
+            True if the amount and currency matches
+            False otherwise
+
+        Raises:
+            DoesNotExist: if there is no order that matches the number given
+
+        """
+        order = Order.objects.get(number=order_num)
+        if order.total_excl_tax == auth_amount and order.currency == auth_currency:
```
clintonb: Simplify: `return order.total_excl_tax == auth_amount and order.currency == auth_currency`

extensions/payment/constants.py
```diff
@@ -32,3 +38,7 @@ class CybersourceConstants(object):
     FIELD_NAMES = CybersourceFieldNames()
     UNSIGNED_FIELD_NAMES = u''
     SEPARATOR = u','
+    CANCEL = u'CANCEL'
+    DECLINE = u'DECLINE'
+    SUCCESS = u'success'
```
clintonb: No capitalization needed?

extensions/payment/exceptions.py
```diff
@@ -0,0 +1,31 @@
+""" Exceptions used by the processor. """
```
clintonb: None of these seem to be used anywhere.

extensions/payment/constants.py
dianakhuang: It is not, for this particular case.

extensions/payment/processors.py
```diff
@@ -73,6 +83,55 @@ def get_transaction_parameters(
...
+
+        Returns:
+            dict
+
+        """
+        try:
+            valid_params = self._verify_signatures(params)
+            if valid_params[CS.FIELD_NAMES.DECISION] == CS.ACCEPT:
+                # make sure the auth amount and currency is what we expect from
+                # Cybersource
```
clintonb: Minor: This comment can be on a single line.

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
...
+        # Validate the signature to ensure that the message is from CyberSource
+        # and has not been tampered with.
+        signed_fields = params.get(CS.FIELD_NAMES.SIGNED_FIELD_NAMES, u'').split(CS.SEPARATOR)
+        data = CS.SEPARATOR.join([u"{0}={1}".format(k, params.get(k, '')) for k in signed_fields])
+        returned_sig = params.get(CS.FIELD_NAMES.SIGNATURE, '')
+        secret = getattr(settings, 'CYBERSOURCE_SECRET_KEY')
+        if sign(data, secret) != returned_sig:
+            raise ProcessorSignatureException()
+
+        # Validate that we have the paramters we expect and can convert them
```
clintonb: s/paramters/parameters

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
...
+        required_params = [
+            (CS.FIELD_NAMES.REQ_REFERENCE_NUMBER, str),
+            (CS.FIELD_NAMES.REQ_CURRENCY, str),
+            (CS.FIELD_NAMES.DECISION, str),
+            (CS.FIELD_NAMES.AUTH_AMOUNT, Decimal),
+        ]
+        for key, key_type in required_params:
+            if key not in params:
+                raise ProcessorDataException(
+                    _(
```
clintonb: This does not need to be translated. The user should never see this message.

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
...
+                        u"The payment processor did not return a required parameter: {parameter}"
+                    ).format(parameter=key)
+                )
+            try:
+                valid_params[key] = key_type(params[key])
+            except (ValueError, TypeError, InvalidOperation):
+                raise ProcessorDataException(
```
clintonb: This does not need to be translated. The user should never see this message.

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
...
+                    _(
```
clintonb: This does not need to be translated. The user should never see this message.

extensions/payment/tests/test_processors.py
```diff
@@ -128,6 +139,120 @@ def test_excessive_merchant_defined_data(self):
...
+        self.assertEqual(result['order'], self.order.number)
+
+    def test_process_payment_rejected(self):
+        """ Simulate a callback from CyberSource indicating that the payment was rejected. """
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total, decision='REJECT', currency=self.order.currency
+        )
+        result = Cybersource().process_processor_response(params)
+
+        # Expect that we get an error message
```
clintonb: Should the error message be tested here as well?

dianakhuang: We don't actually pass the error message all the way back up in the current implementation, but maybe we should. I'll add another parameter in the `result` oject

extensions/payment/tests/test_processors.py
```diff
@@ -128,6 +139,120 @@ def test_excessive_merchant_defined_data(self):
...
+    def test_process_no_credit_card_digits(self):
+        """ Simulate a credit card number with no digits provided """
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total,
+            card_number='nodigits', currency=self.order.currency
+        )
+        result = Cybersource().process_processor_response(params)
+
+        # Expect that we processed the payment successfully
+        self.assertTrue(
```
clintonb: Why would payment be successful in this case?

extensions/payment/processors.py
```diff
@@ -73,6 +83,55 @@ def get_transaction_parameters(
...
+
+        Concrete implementations should:
+            1) Verify the parameters and determine if the payment was successful.
+            2) If successful, mark the order as purchased and send order to fulfillment
+            3) If unsuccessful, try to figure out why and log a helpful error message.
+            4) Return a dictionary of the form:
+                {'success': bool, 'order': Order}
+
+        Args:
+            params (dict): Dictionary of parameters received from the payment processor.
```
stephensanchez: Nitpick, may be worth outlining the expected key/value pairs in the params dict

extensions/payment/tests/test_processors.py
clintonb: As long as its logged, it doesn't have to be passed. I was more confused by the comment. If the message isn't needed, no need to modify your implementation. Just update the test comments.

extensions/payment/processors.py
```diff
@@ -73,6 +83,55 @@ def get_transaction_parameters(
...
+                    CS.FIELD_NAMES.ORDER: valid_params[CS.FIELD_NAMES.REQ_REFERENCE_NUMBER],
+                }
+            else:
+                return {
+                    CS.SUCCESS: False,
+                    CS.FIELD_NAMES.ORDER: valid_params[CS.FIELD_NAMES.REQ_REFERENCE_NUMBER],
+                }
+        except CybersourceError:
+            logger.exception('error processing CyberSource postpay callback')
+            logger.info(json.dumps(params))
```
stephensanchez: Why info for the params, instead of adding that to the same exception log message on line 128?

extensions/payment/tests/test_processors.py
dianakhuang: We don't actually care about the credit card number on our end. Cybersource is the one handling it. I suspect that this test was originally written after we accidentally tried to coerce that field into numbers or something like that.

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
         )
 
         return parameters
+
+    def _check_payment_consistency(self, order, auth_amount, currency):
+        """
+        After the payment has successfully been completed, we need to verify that the
```
stephensanchez: docstring got cut off

extensions/payment/processors.py
dianakhuang: It's a really really long list, unfortunately. Cybersource passes through a lot.

extensions/payment/processors.py
dianakhuang: `logger.exception` is a specific function that will also log the stacktrace of the exception, so I think it makes sense to leave this as a separate line because the params dict tends to be quite large.

extensions/payment/processors.py
```diff
@@ -73,6 +83,55 @@ def get_transaction_parameters(
...
+            valid_params = self._verify_signatures(params)
+            if valid_params[CS.FIELD_NAMES.DECISION] == CS.ACCEPT:
+                # make sure the auth amount and currency is what we expect from
+                # Cybersource
+                self._check_payment_consistency(
+                    valid_params[CS.FIELD_NAMES.REQ_REFERENCE_NUMBER],
+                    valid_params[CS.FIELD_NAMES.AUTH_AMOUNT],
+                    valid_params[CS.FIELD_NAMES.REQ_CURRENCY]
+                )
+                return {
```
stephensanchez: nitpick, but you may be able to simplify the function by creating the response data at the top 
```
result = {
    CS.SUCCESS: False,
    CS.FIELD_NAMES.ORDER: None
}
```
Update CS.FIELD_NAMES.ORDER right after calling _verify_signatures()
and CS.SUCCESS after checking the valid_params dict.
Then return "results" in a finally block.

Not necessary though, just a thought to avoid the multiple return statements.

extensions/payment/processors.py
```diff
@@ -200,3 +259,106 @@ def _add_signature(self, parameters):
...
+            ProcessorUserCancelled: The user cancelled the transaction.
+
+            ProcessorUserDeclined: The payment was declined by the user.
+
+        """
+
+        # First see if the user cancelled the transaction
+        # if so, then not all parameters will be passed back so we can't yet verify signatures
+        if params.get(CS.FIELD_NAMES.DECISION) == CS.CANCEL:
+            raise ProcessorUserCancelled()
```
stephensanchez: The order pipeline has states like "Payment Cancelled" which we could set, what do you think? The problem is that we may be able to set it here, but if the user just closes a window, I'm not sure if CyberSource would send us anything back.

dianakhuang: Closing the window shouldn't trigger CyberSource to send us anything, that's true. But I do think you're right that it makes sense to try to make the state of the order reflect what the user did if we do have that information.

stephensanchez:
Looks good to me. I'm curious what you think about updating the order status based on the payment processor? If we should leverage the payment cancelled state, or create anything new, or give more insight into what's returned?

extensions/payment/processors.py
clintonb: It makes sense to me in this case and the case of the user declining to update the order state.

extensions/payment/tests/test_processors.py
```diff
@@ -177,3 +326,153 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+        Returns:
+            string
+
+        """
+        secret = getattr(settings, "CYBERSOURCE_SECRET_KEY")
+        message = CS.SEPARATOR.join([
+            u"{0}={1}".format(signed_field, params[signed_field])
+            for signed_field
+            in params['signed_field_names'].split(CS.SEPARATOR)
+        ])
```
rlucioni: The logic present in this private method closely mirrors logic embedded in [`_assert_order_parameters`](https://github.com/edx/django-oscar-extensions/pull/35/files#diff-f6d2a8185a50f8da7aa107c1bbf6b530R319). If you need to use the logic on its own elsewhere in your tests, pulling it out is fine, but remember to pull it out of `_assert_order_parameters`.

Also, two notes. You should use `CS.MESSAGE_SUBSTRUCTURE` instead of the raw string `u"{0}={1}"` - [`_assert_order_parameters`](https://github.com/edx/django-oscar-extensions/pull/35/files#diff-f6d2a8185a50f8da7aa107c1bbf6b530R319) serves as an example. I also don't think the note in the docstring associated with this method is relevant.

extensions/payment/tests/test_processors.py
```diff
@@ -177,3 +326,153 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+            "message": "Request was processed successfully.",
+            "auth_response": "100",
+            "req_profile_id": "0000001",
+            "req_transaction_uuid": "ddd9935b82dd403f9aa4ba6ecf021b1f",
+            "auth_trans_ref_no": "85080648RYI23S6I",
+            "req_bill_to_surname": "Doe",
+            "req_bill_to_forename": first_name,
+            "req_bill_to_email": "john@example.com",
+            "req_override_custom_receipt_page": "http://localhost:8000/shoppingcart/postpay_callback/",
+            "req_access_key": "abcd12345",
```
rlucioni: This has to be the nittiest of style nitpicks, but I noticed it so I'm just going to mention it: we're pretty careful about sticking to using single quotes and reserving double-quotes for strings leaving the system to be printed or used as copy. Would you mind adhering to that convention throughout the new code you're adding? It makes things a little more consistent.

extensions/payment/tests/test_processors.py
```diff
@@ -69,13 +73,21 @@ def test_base_processor_unusable(self):
...
+
 
+@ddt.ddt
 class CybersourceTests(PaymentProcessorTestCase):
     """Tests of the CyberSource processor class."""
     UUID_HEX = 'madrigal'
     PI_DAY = datetime.datetime(2015, 3, 14, 9, 26, 53)
     RECEIPT_PAGE_URL = CANCEL_PAGE_URL = 'http://www.lospolloshermanos.com'
     MERCHANT_DEFINED_DATA = [u'ℂ𝕙𝕚𝕝𝕖', u'𝕄𝕖𝕩𝕚𝕔𝕠', u'ℕ𝕖𝕨 𝕄𝕖𝕩𝕚𝕔𝕠']
+    FAILED_DECISIONS = ["DECLINE", "CANCEL", "ERROR"]
```
rlucioni: You've included two of these three status strings in `CybersourceConstants`. If you include `u'ERROR'` there as well, you shouldn't need to write out the strings again here. You can just do `[CS.DECLINE, CS.CANCEL, CS.ERROR]`.

extensions/payment/processors.py
```diff
@@ -73,6 +83,46 @@ def get_transaction_parameters(
 
         return signed_transaction_parameters
 
+    def process_processor_response(self, params):
+
+        """
+        Handle a response from the payment processor.
+
+        Concrete implementations should:
```
rlucioni: Is this not a concrete implementation?

extensions/payment/processors.py
```diff
@@ -28,6 +34,10 @@ def get_transaction_parameters(
         """Generate a dictionary of transaction parameters to be sent to a payment processor."""
         raise NotImplementedError("Transaction parameters method not implemented.")
 
+    def process_processor_response(self, params):
```
rlucioni: Again a nitpick (sorry!), but I think `handle_processor_response` would be a better name; that alliteration is killer!

extensions/payment/tests/test_processors.py
rlucioni: Are "digits" strings or integers? Should this test fail if `card_number` is an integer?

extensions/payment/tests/test_processors.py
```diff
@@ -128,6 +140,143 @@ def test_excessive_merchant_defined_data(self):
...
+    def test_process_payment_rejected(self):
+        """ Simulate a callback from CyberSource indicating that the payment was rejected. """
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total, decision='REJECT', currency=self.order.currency
+        )
+        result = Cybersource().process_processor_response(params)
+
+        # Expect that we get an error
+        self.assertFalse(result['success'])
+        self.assertEqual(result['order'], self.order.number)
```
rlucioni: It looks like these tests were pulled over from edx-platform. Since you've included keys like `'success'` and `'order'` in `CybersourceConstants` as well as various status strings, you should use those constants.

extensions/payment/constants.py
```diff
@@ -19,6 +21,10 @@ class CybersourceFieldNames(object):
     SIGNED_FIELD_NAMES = u'signed_field_names'
     UNSIGNED_FIELD_NAMES = u'unsigned_field_names'
     SIGNATURE = u'signature'
+    ACCEPTED = u'accepted'
+    AUTH_AMOUNT = u'auth_amount'
+    DECISION = u'decision'
+    ORDER = u'order'
```
rlucioni: Judging from how you're using this `ORDER` constant, I'd say it doesn't belong in the object, but instead the one below. This object is only intended to hold "CyberSource request field names" (i.e., names of parameters returned to us by CyberSource).

extensions/payment/processors.py
```diff
@@ -200,3 +250,115 @@ def _add_signature(self, parameters):
...
+        # if so, then auth_amount will not be passed back so we can't yet verify signatures
+        if params.get(CS.FIELD_NAMES.DECISION) == CS.DECLINE:
+            self._mark_cancelled(params.get(CS.FIELD_NAMES.REQ_REFERENCE_NUMBER))
+            raise ProcessorUserDeclined()
+
+        # Validate the signature to ensure that the message is from CyberSource
+        # and has not been tampered with.
+        signed_fields = params.get(CS.FIELD_NAMES.SIGNED_FIELD_NAMES, u'').split(CS.SEPARATOR)
+        data = CS.SEPARATOR.join([u"{0}={1}".format(k, params.get(k, '')) for k in signed_fields])
+        returned_sig = params.get(CS.FIELD_NAMES.SIGNATURE, '')
```
rlucioni: I cleaned up and abstracted this logic in [`_add_signature`](https://github.com/edx/django-oscar-extensions/pull/35/files#diff-a2872756ad4ab375895e1abb60fafdf3R217), which includes a docstring explaining why this is being done. I'd suggest using it instead of duplicating the logic here.

extensions/payment/processors.py
```diff
@@ -200,3 +250,115 @@ def _add_signature(self, parameters):
...
+        # Validate that we have the parameters we expect and can convert them
+        # to the appropriate types.
+        # Usually validating the signature is sufficient to validate that these
+        # fields exist, but since we're relying on CyberSource to tell us
+        # which fields they included in the signature, we need to be careful.
+        valid_params = {}
+        required_params = [
+            (CS.FIELD_NAMES.REQ_REFERENCE_NUMBER, str),
+            (CS.FIELD_NAMES.REQ_CURRENCY, str),
+            (CS.FIELD_NAMES.DECISION, str),
```
rlucioni: Since the values associated with these keys are expected to be text, it would be more appropriate (and potentially safer) to convert them using `unicode` instead of `str`.

extensions/payment/tests/test_processors.py
dianakhuang: The 'digits' will always come to us as a string. Sometimes they will have digits in them. Sometimes they will not. This test is to ensure that we do not fail if we get something that does not have digits in them.

dianakhuang:
I believe I've addressed all code review comments. @clintonb @rlucioni @stephensanchez

stephensanchez:
:+1:

extensions/payment/constants.py
```diff
@@ -19,6 +21,9 @@ class CybersourceFieldNames(object):
     SIGNED_FIELD_NAMES = u'signed_field_names'
     UNSIGNED_FIELD_NAMES = u'unsigned_field_names'
     SIGNATURE = u'signature'
+    ACCEPTED = u'accepted'
```
rlucioni: It doesn't look like you use this constant anywhere.

extensions/payment/processors.py
```diff
@@ -1,14 +1,21 @@
 """Payment processing classes containing logic specific to particular payment processors."""
 import uuid
+import json
 import logging
 import datetime
+
 from collections import OrderedDict
+from decimal import Decimal, InvalidOperation
```
rlucioni: Nit: these are both standard library imports and shouldn't be separated from the rest of the standard library imports with a newline ([reference](https://www.python.org/dev/peps/pep-0008/#imports)).

extensions/payment/errors.py
```diff
@@ -14,3 +14,28 @@ class CybersourceError(PaymentProcessorError):
...
+    """ The user did not pay the correct amount. """
+    pass
+
+
+class ProcessorUserCancelled(CybersourceError):
+    """ The user cancelled the transaction. """
+    pass
+
+
+class ProcessorUserDeclined(CybersourceError):
```
rlucioni: Prefixing these exception names with "Processor" makes them sound like standard payment processor errors. It's also clear that they're processor errors, since they subclass `PaymentProcessorError`. I'd remove the "Processor" prefix from the names.

extensions/payment/processors.py
```diff
@@ -69,9 +80,52 @@ def get_transaction_parameters(
...
+            if valid_params[CS.FIELD_NAMES.DECISION] == CS.ACCEPT:
+                # make sure the auth amount and currency is what we expect from Cybersource
+                self._check_payment_consistency(
+                    valid_params[CS.FIELD_NAMES.REQ_REFERENCE_NUMBER],
+                    valid_params[CS.FIELD_NAMES.AUTH_AMOUNT],
+                    valid_params[CS.FIELD_NAMES.REQ_CURRENCY]
+                )
+                result['success'] = True
+        except CybersourceError:
+            logger.exception('error processing CyberSource postpay callback')
```
rlucioni: "Postpay callback" was the name in the old system. It would be clearer to log:
```python
logger.exception(u"Error handling CyberSource response.")
```

extensions/payment/processors.py
```diff
@@ -192,11 +249,119 @@ def _add_signature(self, parameters):
...
+        After the payment has successfully been completed, we need to verify that the authorized amount and the currency
+        match what we expect based on the order number.
 
-        return parameters
+        Args:
+            order (str): order number of the order being processed
+            auth_amount (num): the amount being paid
+
+        Raises:
+            ProcessorWrongAmountException: indicates that the processor
```
rlucioni: Indicates that the processor what? I need to know what happens next! Such suspense!

extensions/payment/processors.py
```diff
@@ -192,11 +249,119 @@ def _add_signature(self, parameters):
...
+
+        Returns:
+            nothing
```
rlucioni: `None`.

extensions/payment/processors.py
```diff
@@ -192,11 +249,119 @@ def _add_signature(self, parameters):
...
+        return valid_params
+
+    def _mark_cancelled(self, order_num):
+        """ When a user cancels a submitted payment, mark it as such in the DB """
+        try:
+            order = Order.objects.get(number=order_num)
+            order.set_status(ORDER.ORDER_CANCELLED)
+        except Order.DoesNotExist:
+            raise ProcessorDataException(
+                "The payment processor accepted an order with number {number} that is not in our system."
```
rlucioni: For clarity, most of the log messages already in place use brackets to demarcate interpolated values. Could you modify these log messages so that they're consistent with that convention?

extensions/payment/tests/test_processors.py
```diff
@@ -69,16 +74,52 @@ def test_base_processor_unusable(self):
         """Test that use of the base payment processor class fails due to NotImplementedError."""
         BasePaymentProcessor().get_transaction_parameters(self.order)
 
+    @raises(NotImplementedError)
+    def test_processor_reponse_unusable(self):
+        """ Test that we can't use the base processor_response """
```
rlucioni: This could be folded into `test_base_processor_unusable` above.

dianakhuang: Yes, but you'd still need to check that each call raises the NotImplementedError separately anyway. I don't see a huge difference either way.

extensions/payment/tests/test_processors.py
```diff
@@ -69,16 +74,52 @@ def test_base_processor_unusable(self):
...
-    """Tests of the CyberSource processor class."""
+    """ Abstract test class for Cybersource tests. """
+
+    def _signature(self, params):
+        """
+        Calculate the signature from a dictionary of params.
+
+        NOTE: This method uses the processor's hashing method.  That method
+        is a thin wrapper of standard library calls, and it seemed overly complex
+        to rewrite that code in the test suite.
```
rlucioni: > ...it seemed overly complex to rewrite that code in the test suite.

This function itself is a rewritten version of code in the actual CyberSource processor class, `_generate_signature`. Making that method public and using it in your tests would be cleaner than repeating it here.

extensions/payment/processors.py
```diff
@@ -69,9 +80,52 @@ def get_transaction_parameters(
...
+            params (dict): Dictionary of parameters received from the payment processor.
+
+        Keyword Args:
+            Can be used to provide additional information to concrete implementations.
+
+        Returns:
+            dict
+
+        """
+        result = {'success': False, 'order': None}
```
rlucioni: Why not wrap these keys up in CybersourceConstants? They're not field names, but they're still constants important to the processor implementation.

dianakhuang: They aren't CyberSource-specific constants, though.

extensions/payment/tests/test_processors.py
dianakhuang: I'm not comfortable making the method public just because we want to use it in tests. I've updated the functions that use `_signature` to use `_generate_signature` instead.

extensions/payment/processors.py
rlucioni: The `ISO_8601_FORMAT` and `SEPARATOR` (a comma) aren't really "CyberSource specific" either, but they're constants important to the processor class implementation that we'd like to have access to elsewhere. The idea behind `CybersourceConstants` is to prevent tests and other areas that might need access to these constants from breaking if the values change. Since the 'success' and 'order' strings *are* constant and critical to the implementation of this processor class, they should be assigned to constant variables.

extensions/payment/tests/test_processors.py
rlucioni: You're right that there isn't a huge difference. However, If you're going to test them separately, the names and comments for the tests should at least be modified, for the sake of clarity. The name and docstring for `test_base_processor_unusable` makes it sound like it's responsible for checking that the entire abstract processor class is unusable, when in reality it only checks `get_transaction_parameters`.

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +182,278 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+            'req_bill_to_surname': 'Doe',
+            'req_bill_to_email': 'john@example.com',
+            'req_override_custom_receipt_page': 'http://localhost:8000/shoppingcart/postpay_callback/',
+            'req_access_key': 'abcd12345',
+        }
+
+        # if decision is in FAILED_DECISIONS list then remove the auth_amount from params dict
+
+        if decision in self.FAILED_DECISIONS:
+            del params['auth_amount']
```
rlucioni: This string (and 'signature', below) are in `CybersourceConstants`; you should use them!

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +183,287 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
-        expected_parameters[CS.FIELD_NAMES.SIGNATURE] = sign(message, secret)
+        # Expect that we processed the payment successfully
+        self.assertTrue(result[PC.SUCCESS])
+        self.assertEqual(result[PC.ORDER], self.order.number)
 
-        self.assertEqual(returned_parameters, expected_parameters)
+    def test_process_payment_rejected(self):
+        """ Simulate a callback from CyberSource indicating that the payment was rejected. """
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total, decision='REJECT', currency=self.order.currency
```
rlucioni: Is `REJECT` a standard CyberSource error status/decision? If so, it should be in `CybersourceConstants`.

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +183,287 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+
+        # Expect an error
+        self.assertFalse(result[PC.SUCCESS])
+
+    def test_process_user_declined(self):
+        """ Simulate a user declining the transaction """
+        # set the order status to what we expect after being sent out to CyberSource
+        self.order.set_status(ORDER.BEING_PROCESSED)
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total, currency=self.order.currency, decision=u'DECLINE'
```
rlucioni: `DECLINE` is available in `CybersourceConstants`, as are other decision strings in the lines above this one.

extensions/payment/tests/test_processors.py
dianakhuang: It turns out that REJECT is no longer one of the decisions that CyberSource will send back to us, so I removed the test.

dianakhuang:
All good, @rlucioni ?

clintonb:
:+1:

extensions/payment/processors.py
```diff
@@ -192,11 +249,120 @@ def _add_signature(self, parameters):
...
+            None
+        """
+        try:
+            if not Order.check_order_total(order, auth_amount, currency):
+                raise WrongAmountException(
+                    u"The amount charged by the processor [{charged_amount}] [{charged_amount_currency}] is different "
+                    u"than the total cost of the order for order [{order_number}].".format(
+                        charged_amount=auth_amount,
+                        charged_amount_currency=currency,
+                        order_number=order
```
rlucioni: It looks like you're trying to log an order number here, and below on line 284. You should use `order.number`, not `order`.

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +183,277 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+                              'auth_trans_ref_no',
+                              'auth_time',
+                              'bill_trans_ref_no',
+                              'signed_field_names',
+                              'signed_date_time']
+
+        # if decision is in FAILED_DECISIONS list then remove  auth_amount from
+        # signed_field_names list.
+        if decision in self.FAILED_DECISIONS:
+            signed_field_names.remove('auth_amount')
```
rlucioni: The "auth_amount" field name is available in `CybersourceConstants`.

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +183,277 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+        params = self._signed_callback_params(
+            self.order.number, self.order_total, self.order_total, currency=self.order.currency
+        )
+        del params[missing_param]
+
+        # Recalculate the signature with no signed fields so we can get past
+        # signature validation.
+        params['signed_field_names'] = 'reason_code,message'
+        # pylint: disable=protected-access
+        params['signature'] = Cybersource()._generate_signature(params)
```
rlucioni: These field names are also available in `CybersourceConstants`.

extensions/payment/processors.py
dianakhuang: It's correct in this method, but not on line 367.

extensions/payment/tests/test_processors.py
```diff
@@ -167,13 +183,277 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
...
+        params[CS.FIELD_NAMES.SIGNATURE] = Cybersource()._generate_signature(params)
+
+        result = Cybersource().handle_processor_response(params)
+
+        # Expect an error
+        self.assertFalse(result[PC.SUCCESS])
+
+    def _signed_callback_params(
+            self, order_id, order_amount, paid_amount,
+            decision='ACCEPT', signature=None, card_number='xxxxxxxxxxxx1111',
```
rlucioni: 'ACCEPT' is available in `CybersourceConstants`.

rlucioni:
One last note, then :+1: one the build passes.

PR #36, opened 2015-03-26
Enable ATOMIC_REQUESTS for local development

rlucioni:
Wraps views in a transaction to more closely mirror production config, and removes what would be a redundant `atomic()` context manager. Connection pooling is also used in production, but Django docs discourage its use in development.

Previously reviewed in #34.


clintonb:
:+1:

PR #37, opened 2015-03-26
Use newer request `data` attribute instead of the old-style `DATA` attribute

rlucioni:
Old-style attributes like `DATA` and `FILES` are pending deprecation in favor of the unified `data` attribute ([reference](http://www.django-rest-framework.org/api-guide/requests/#data-and-files)).

@clintonb or @stephensanchez, could I get a thumb?


clintonb:
:+1:

PR #38, opened 2015-03-26
XCOM-128: Create an endpoint for getting orders.

stephensanchez:
Simple getter endpoint for orders. 
https://openedx.atlassian.net/browse/XCOM-128

Basic logic in the endpoint is to ensure the user in the request matches the user associated with the order, and that the order is in some state of "PAID" or beyond. 

Renamed the view for creating orders (CreateOrderView) based on a POST with associated SKUs. 

Review when able. There will be an associated PR for Platform to call this endpoint. 
@rlucioni @dianakhuang @clintonb


extensions/api/urls.py
```diff
@@ -5,6 +5,7 @@
 
 urlpatterns = patterns(
     '',
-    url(r'orders/$', views.OrdersView.as_view(), name='orders'),
+    url(r'orders/(?P<number>[-\w]+)$', views.OrdersView.as_view(), name='orders'),
```
rlucioni: Could the name for this URL be more descriptive of its intended use? Maybe `get_order`?

extensions/api/views.py
```diff
@@ -42,7 +43,68 @@ def _fulfill_order(self, order):
...
+                "total_excl_tax": 0.0
+            }'
+        """
+        order = self.get_object()
+        if order.is_paid and order.user == request.user:
+            return Response(self.get_serializer(order).data)
+        return Response(status=status.HTTP_404_NOT_FOUND)
+
+
+class CreateOrderView(FulfillmentMixin, APIView):
```
rlucioni: What prevents the `get` and `post` views from being part of the same `OrdersView` class? I'm unclear on why this separation is necessary.

extensions/api/views.py
```diff
@@ -42,7 +43,68 @@ def _fulfill_order(self, order):
...
+    serializer_class = serializers.OrderSerializer
+
+    lookup_field = 'number'
+    queryset = Order.objects.all()
+
+    def get(self, request, *args, **kwargs):  # pylint: disable=unused-argument
+        """Allow the viewing of Paid Orders.
+
+        Given an order number, allow the viewing of a paid order. This endpoint will only return an order if
+        it in a PAID state, or a proceeding state in the order workflow (COMPLETE, FULFILLMENT_ERROR, REFUNDED). This
```
rlucioni: s/it in/it is in

extensions/api/urls.py
clintonb: Add a trailing slash.

clintonb: Create a constant for `(?P<number>[-\w]+)`.

extensions/api/views.py
```diff
@@ -42,7 +43,68 @@ def _fulfill_order(self, order):
         return fulfilled_order
 
 
-class OrdersView(FulfillmentMixin, APIView):
+class OrdersView(GenericAPIView):
```
clintonb: This can be simplified: 
1. Inherit from `RetrieveAPIView`.
2. Set `allow_empty` to `False`.
3. Override `get_object` to check the order state and user, and return None if the state/user are invalid.

stephensanchez:
Thanks for the review guys, code is updated.

extensions/api/views.py
stephensanchez: I had an offline conversation with Clinton because I could see this going either way. The original intent I had was to split them because the ```get``` class was directly interacting with orders, whereas the ```post``` class view dealt with SKUs to create orders. 

I'd prefer to keep them separate for now. Clinton also recommended some nice generic DRF view for me to extend to simplify my class.

stephensanchez:
@clintonb updated per comment, but I could not override ```get_object()``` since it does not have access to the ```request.user```, which I want to check before returning the object.   I did use ```allow_empty``` though, since that's a nice check.

extensions/api/views.py
```diff
@@ -42,7 +42,68 @@ def _fulfill_order(self, order):
...
 
 
-class OrdersView(FulfillmentMixin, APIView):
+class RetrieveOrderView(RetrieveAPIView):
+    """Endpoint for viewing an order."""
+    throttle_classes = (OrdersThrottle,)
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+    serializer_class = serializers.OrderSerializer
+    allow_empty = False
```
clintonb: My fault for screwing up my comment update. DRF is deprecating `allow_empty`. Instead, your code should raise `Http404` when the check fails. Also, if you want to try overriding `get_object`, you should have access to `self.request`.

stephensanchez: ah, didn't realize the class itself had the request as an attribute. I'll tweak that.

extensions/api/views.py
clintonb: Follow-up: Views that act on the entire order space (`/orders/`) can be grouped. This includes listing and creating. Views that act on individual orders (`/orders/:id/`) can be grouped. These include order retrieval and updating--not deleting, since we *never* delete orders. Grouping both of these groups causes issues in the case of the `GET` action. A `GET` on `/orders/` is not the same as a `GET` on `/orders/:id/`

stephensanchez:
@clintonb Now the view overrides ```get_object()``` as suggested!

stephensanchez:
Coveralls seems stuck, but 100% coverage
https://coveralls.io/builds/2201848

extensions/api/urls.py
```diff
@@ -2,9 +2,19 @@
...
 from extensions.api import views
 
+ORDER_NUMBER_PATTERN = "(?P<number>[-\w]+)"  # pylint: disable=anomalous-backslash-in-string
 
 urlpatterns = patterns(
     '',
-    url(r'orders/$', views.OrdersView.as_view(), name='orders'),
-    url(r'orders/(?P<number>[-\w]+)/fulfill/$', views.FulfillOrderView.as_view(), name='fulfill_order'),
+    url(
+        r'orders/{number}$'.format(number=ORDER_NUMBER_PATTERN),
```
clintonb: Add a trailing slash.

extensions/api/views.py
```diff
@@ -41,7 +42,60 @@ def _fulfill_order(self, order):
...
+    """Endpoint for viewing an order."""
+    throttle_classes = (OrdersThrottle,)
+    authentication_classes = (JwtAuthentication,)
+    permission_classes = (IsAuthenticated,)
+    serializer_class = serializers.OrderSerializer
+    lookup_field = 'number'
+    queryset = Order.objects.all()
+
+    def get_object(self):
+        """Allow the viewing of Paid Orders.
```
clintonb: [NTH] Move this comment to the view class, instead of the method.

clintonb:
:+1:

extensions/api/urls.py
```diff
@@ -2,9 +2,19 @@
 
 from extensions.api import views
 
+ORDER_NUMBER_PATTERN = "(?P<number>[-\w]+)"  # pylint: disable=anomalous-backslash-in-string
```
rlucioni: Instead of telling pylint to ignore this string, use Python's raw string notation. It's intended to be used for regular expression patterns.
```python
ORDER_NUMBER_PATTERN = r'(?P<number>[-\w]+)'
```

stephensanchez:
@rlucioni updated the string.

rlucioni:
:+1:

PR #39, opened 2015-03-27
Change order numbers to have a set prefix

stephensanchez:
Order numbers have a set prefix now. Out of the box, I chose to have the prefix be 'OSCR', but I imagine we'll want to change this to 'edX' for production.

@rlucioni @dianakhuang @clintonb 

FYI @feanil if we want a different prefix in production, I imagine I'll add this to the configuration list we already have in place, should I also create a devops ticket?


rlucioni:
I'd recommend "EDX" as the production prefix, to be consistent with the all caps "OSCR" prefix. The service settings page includes a note that any new settings should be accompanied by a DevOps ticket, so I think it would be a good idea to file one for this.

extensions/order/tests/test_utils.py
```diff
@@ -0,0 +1,20 @@
...
+from unittest import TestCase
+from django.test import override_settings
+from oscar.test.newfactories import BasketFactory
+from extensions.order.utils import OrderNumberGenerator
+
+
+class UtilsTest(TestCase):
+    """Unit tests for the order utility functions and classes. """
+
+    @override_settings(ORDER_NUMBER_PREFIX="Zoidberg")
```
clintonb: Minor: Make `"Zoidberg"` a constant.

clintonb:
:+1:

extensions/order/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+        Args:
+            basket (Basket): Used to construct the order ID.
+
+        Returns:
+            String: representation of the order 'number' with a configured prefix.
+
+        """
+        prefix = getattr(settings, 'ORDER_NUMBER_PREFIX', 'OSCR')
+        order_id = str(100000 + basket.id)
```
rlucioni: We spoke offline about this leaking information about the number of baskets in our system, but I'll just add a note here so that we remember to talk about the resolution.

clintonb:
@stephensanchez Would you also add a test to ensure the generate order numbers are matched by the order number regex in `api/urls.py`?

rlucioni:
This is relatively minor, but the commit message you're using no longer reflects the changes being made.

stephensanchez:
@clintonb @rlucioni I've made the suggested changes.

clintonb:
:+1:  (again)

extensions/order/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+    ORDER_NUMBER_PREFIX = "Zoidberg"
+
+    @override_settings(ORDER_NUMBER_PREFIX=ORDER_NUMBER_PREFIX)
+    def create_order_number(self):
+        """Test creating order numbers"""
+        basket = BasketFactory()
+        next_basket = BasketFactory()
+        new_order_number = OrderNumberGenerator().order_number(basket)
+        next_order_number = OrderNumberGenerator.order_number(next_basket)
```
rlucioni: Since `order_number` is a static method, you should be able to call it everywhere like you do here. There shouldn't be a need to create an instance of the `OrderNumberGenerator` class.

extensions/api/views.py
```diff
@@ -252,8 +252,7 @@ def _prepare_order(self, basket, product, sku, payment_processor):
             shipping_method,
             shipping_charge,
             user=basket.owner,
-            # OrderNumberGenerator returns integers, yet order numbers should be strings.
-            order_number=unicode(OrderNumberGenerator().order_number(basket)),
+            order_number=OrderNumberGenerator().order_number(basket),
```
rlucioni: You missed one!

rlucioni:
:+1:

PR #40, opened 2015-03-27
Disable default Oscar analytics receivers

rlucioni:
Some default receivers, such as the receiver responsible for tallying product orders, make row-locking queries which significantly degrade performance at scale. This PR introduces a feature flag which controls whether these receivers should be installed at startup.

@clintonb and @wedaly, please review.


rlucioni:
I've updated the [service settings page](https://openedx.atlassian.net/wiki/display/XCOM/Service+Settings) and filed [a ticket](https://openedx.atlassian.net/browse/DEVOPS-1435) with DevOps.

clintonb:
:+1:

rlucioni:
@wedaly has deferred review of this PR; I'm going to go ahead and merge.

PR #41, opened 2015-03-27
Add processor response endpoint

dianakhuang:
XCOM-154

This is the simplistic endpoint that CyberSource contacts in order to update the order information. It will then record the payment, update the order information, and then attempt to fulfill the order.

I have also updated our CyberSource processor to update the order status when payment fails. More sophisticated error handling will happen as part of XCOM-174.

@rlucioni @clintonb @stephensanchez


extensions/fulfillment/mixins.py
```diff
@@ -0,0 +1,20 @@
...
+
+EventHandler = get_class('order.processing', 'EventHandler')
+
+
+class FulfillmentMixin(object):
+    """ A mixin that provides the ability to fulfill orders. """
+    SHIPPING_EVENT_NAME = 'Shipped'
+
+    def _fulfill_order(self, order):
+        """Attempt fulfillment for orders consisting of a free product."""
```
rlucioni: Since this method isn't specific to free products/orders, this docstring should be updated to `"Attempt fulfillment for an order."`

extensions/payment/app.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class PaymentApplication(Application):
+    """Payment application class.
+
+    This subclasses Oscar's base application class to create a custom
+    container for Payment URLs and views
+    """
+    def get_urls(self):
+        """Returns the URL patterns for the API."""
```
rlucioni: `"""Returns the URL patterns for the payment app."""`

extensions/payment/processors.py
```diff
@@ -356,11 +361,11 @@ def _verify_signatures(self, params):
 
         return valid_params
 
-    def _mark_cancelled(self, order_num):
+    def _mark_status(self, order_num, status):
         """ When a user cancels a submitted payment, mark it as such in the DB """
```
rlucioni: Could you update this docstring to reflect the more general functionality this method now provides?

extensions/payment/processors.py
```diff
@@ -269,6 +269,7 @@ def _check_payment_consistency(self, order_num, auth_amount, currency):
         """
         try:
             if not Order.check_order_total(order_num, auth_amount, currency):
+                self._mark_status(order_num, ORDER.PAYMENT_ERROR)
```
rlucioni: These two lines each require the same SELECT operation to look up the order by number. It would be more efficient to attempt the lookup at the outset. If the order exists, assign the order object to a variable and operate on it directly; no need for the `_mark_status` helper method. If the order doesn't exist, raise the appropriate exception.

extensions/payment/constants.py
```diff
@@ -5,6 +5,7 @@ class ProcessorConstants(object):
     """ Constants that are used by all payment processors """
     ORDER = 'order'
     SUCCESS = 'success'
+    PAID = 'Paid'
```
clintonb: [NIT] `paid` with a lowercase `p`.

extensions/payment/processors.py
```diff
@@ -313,15 +314,19 @@ def _verify_signatures(self, params):
         # First see if the user cancelled the transaction
         # if so, then not all parameters will be passed back so we can't yet verify signatures
         if params.get(CS.FIELD_NAMES.DECISION) == CS.CANCEL:
-            self._mark_cancelled(params.get(CS.FIELD_NAMES.REQ_REFERENCE_NUMBER))
+            self._mark_status(params.get(CS.FIELD_NAMES.REQ_REFERENCE_NUMBER), ORDER.PAYMENT_CANCELLED)
             raise UserCancelled()
 
         #  if the user decline the transaction
         # if so, then auth_amount will not be passed back so we can't yet verify signatures
```
clintonb: [NIT] Fix the grammar of this comment.

extensions/payment/processors.py
```diff
@@ -313,15 +314,19 @@ def _verify_signatures(self, params):
         # First see if the user cancelled the transaction
         # if so, then not all parameters will be passed back so we can't yet verify signatures
```
clintonb: [NIT] Fix the grammar of this comment.

extensions/payment/tests/test_processors.py
```diff
@@ -332,6 +332,16 @@ def test_process_missing_parameters(self, missing_param):
         # Expect an error
         self.assertFalse(result[PC.SUCCESS])
 
+    def test_process_error_parameters(self):
+        """ Test when CyberSource returns an ERROR decision """
```
rlucioni: Could you describe the expected behavior when CyberSource returns an error decision?

extensions/payment/tests/test_views.py
```diff
@@ -0,0 +1,144 @@
...
+        self.user = factories.UserFactory(username='edx')
+        self.order = factories.create_order(number=ORDER_NUMBER, status=ORDER.BEING_PROCESSED)
+        # Create a type of event that gets used during fulfillment
+        ShippingEventType.objects.create(code='shipped', name='Shipped')
+        PaymentEventType.objects.create(name=PC.PAID)
+
+    # Reuse the fake fulfillment module provided by the test_api tests
+    @override_settings(
+        PAYMENT_PROCESSOR_CLASS=SUCCESS_PROCESSOR,
+        FULFILLMENT_MODULES=['fulfillment.tests.test_api.FakeFulfillmentModule', ]
```
clintonb: Override the fulfillment module for the entire class instead of repeating the directive for each test.

extensions/payment/views.py
```diff
@@ -0,0 +1,63 @@
...
+    def post(self, request):
+        """ Handle the response we've been given from the processor. """
+        payment_processor = get_processor_class(getattr(settings, 'PAYMENT_PROCESSOR_CLASS'))
+        # check the data we get
+        params = request.POST.dict()
+        result = payment_processor().handle_processor_response(params)
+        if result[PC.SUCCESS]:
+            # get the order
+            order = Order.objects.get(number=result[PC.ORDER])
+            print order
```
rlucioni: Left over from debugging?

extensions/payment/urls.py
```diff
@@ -0,0 +1,9 @@
+""" Payment-related URLs """
+from django.conf.urls import patterns, url
+
+from extensions.payment import views
+
+urlpatterns = patterns(
+    '',
+    url(r'/processor_callback/$', views.ProcessorResponseView.as_view(), name='processor_callback'),
```
clintonb: This should be CyberSource-specific (e.g. `/cybersource/callback/`). Looking forward to other work, we will need a similar endpoint for Paypal, and sharing an endpoint may lead to ugly branching in the view code.

clintonb:
@dianakhuang Everything looks fine, excepting the callback URL. Let me know if you have questions, or if my comment needs additional clarification.

extensions/payment/constants.py
rlucioni: These constant names should be changed to better reflect their use. `ORDER` and `SUCCESS` are dictionary keys. Could you follow the same pattern used for storing CyberSource field names, and modify `ORDER` to `ORDER_NUMBER`, since it keys the order number returned by the processor instead of an order object? Similarly, `PAID` is used as the name of a payment event; could you change it to something like `PAYMENT_EVENT_NAME`?

extensions/payment/tests/test_processors.py
```diff
@@ -194,6 +194,10 @@ class CybersourcePaymentAcceptanceTests(CybersourceTests):
     """Tests of the CyberSource processor class related to checking response."""
     FAILED_DECISIONS = ["DECLINE", "CANCEL", "ERROR"]
 
+    def setUp(self):
+        super(CybersourcePaymentAcceptanceTests, self).setUp()
+        self.order.set_status(ORDER.BEING_PROCESSED)
```
rlucioni: I'd move this to the base `PaymentProcessorTestCase`. Payment processors should never operate on an order in a state other than `BEING_PROCESSED`. A unit test of this behavior might be a good thing to add.

extensions/payment/urls.py
rlucioni: The URL and name of the endpoint should also more precisely reflect its function. In this case, the endpoint is used for order fulfillment. I'd suggest `/cybersource/fulfill/` and `cybersource_fulfill` as the URL and name, respectively.

extensions/payment/views.py
```diff
@@ -0,0 +1,63 @@
+""" Views for interacting with the payment processor. """
+from django.conf import settings
+from django.http import HttpResponse
+from django.views.generic import View
+
+from oscar.apps.checkout.mixins import OrderPlacementMixin
+from oscar.apps.payment.models import SourceType
```
rlucioni: These third-party imports shouldn't be separated from other third-party imports with a newline.

extensions/payment/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class ProcessorResponseView(View, OrderPlacementMixin, FulfillmentMixin):
+    """
+    Accept response from the processor and fulfill the request
+    """
+
+    def post(self, request):
+        """ Handle the response we've been given from the processor. """
+        payment_processor = get_processor_class(getattr(settings, 'PAYMENT_PROCESSOR_CLASS'))
```
rlucioni: This setting will be removed, pending the merge of #42. To get the default payment processor, you should instead use:
```python
payment_processor = get_processor_class(settings.PAYMENT_PROCESSORS[0])
```

extensions/payment/views.py
```diff
@@ -0,0 +1,63 @@
...
+from oscar.apps.payment.models import SourceType
+
+from extensions.order.models import Order
+from extensions.fulfillment.status import ORDER
+from extensions.fulfillment.mixins import FulfillmentMixin
+from extensions.payment.constants import ProcessorConstants as PC
+from extensions.payment.helpers import get_processor_class
+
+
+class ProcessorResponseView(View, OrderPlacementMixin, FulfillmentMixin):
```
rlucioni: This endpoint should be CyberSource specific. You could follow the same pattern used for defining the processor classes.

rlucioni: I'd also recommend using DRF, to be consistent with our other endpoints. You can follow the pattern established by our other API endpoints; for example, the [fulfillment retry endpoint](https://github.com/edx/django-oscar-extensions/pull/41/files#diff-0d2ac1441ae5eee848b90cf82393d79dR281).

extensions/settings/_oscar.py
```diff
@@ -53,7 +53,8 @@
 OSCAR_ORDER_STATUS_PIPELINE = {
     ORDER.OPEN: (ORDER.BEING_PROCESSED, ORDER.ORDER_CANCELLED,),
     ORDER.ORDER_CANCELLED: (),
-    ORDER.BEING_PROCESSED: (ORDER.PAID, ORDER.PAYMENT_CANCELLED,),
+    ORDER.BEING_PROCESSED: (ORDER.PAID, ORDER.PAYMENT_CANCELLED, ORDER.PAYMENT_ERROR),
+    ORDER.PAYMENT_ERROR: (),
```
rlucioni: Is `PAYMENT_ERROR` the correct final state for orders experiencing a payment error? Other error states allow transition to `COMPLETE` or `REFUNDED`. It seems like this state should allow transition to the `PAID` and `PAYMENT_CANCELLED` states.

extensions/urls.py
```diff
@@ -16,6 +17,7 @@
     # Oscar URLs
     url(r'^i18n/', include('django.conf.urls.i18n')),
     url(r'^api/v1/', include(api.urls)),
+    url(r'^payment/', include(payment.urls)),
```
rlucioni: I'm concerned about these payment endpoints existing separately from the rest of the API. Aren't they part of our nascent API? Why not include them at `api/v1/payment/`?

dianakhuang: If we think of an API as an interface we may want other people to program against, I really don't think this fits at all. It's always going to be keyed to the implementation of our processors without any regard for breaking compatibility for the possible people who would want to write code against it. It is an endpoint that interfaces with an API, but I don't think it makes sense as part of an API itself.

clintonb: I agree. This should *not* be part of the API.

rlucioni: Okay, thanks for the clarification.

dianakhuang:
@rlucioni I believe I've addressed all your comments.

extensions/payment/constants.py
```diff
@@ -3,8 +3,9 @@
 
 class ProcessorConstants(object):
     """ Constants that are used by all payment processors """
-    ORDER = 'order'
+    ORDER_NUMBER = 'order_number'
     SUCCESS = 'success'
+    PAID_EVENT_NAME = 'Paid'
```
rlucioni: Minor: `PAYMENT_EVENT_NAME` would be more correct.

dianakhuang: I'm just worried that it'll be confusing with the fact that we're creating `PaymentEventType` obejcts with the constant `PAYMENT_EVENT_NAME`. Currently, it's the only type, but I can imagine us adding more later to track things like refunds and such.

extensions/payment/processors.py
```diff
@@ -329,16 +329,21 @@ def _verify_signatures(self, params):
...
+        # passed back, so we can't yet verify signatures.
         if params.get(CS.FIELD_NAMES.DECISION) == CS.CANCEL:
-            self._mark_cancelled(params.get(CS.FIELD_NAMES.REQ_REFERENCE_NUMBER))
+            self._mark_status(params.get(CS.FIELD_NAMES.REQ_REFERENCE_NUMBER), ORDER.PAYMENT_CANCELLED)
             raise UserCancelled()
 
-        # If the user declines the transaction auth_amount will not be passed back so we can't yet verify signatures.
+        # If the processor declines the transaction, the auth_amount will
+        # not be passed back so we can't yet verify signatures.
         if params.get(CS.FIELD_NAMES.DECISION) == CS.DECLINE:
```
rlucioni: It seems like CANCEL and DECLINE decisions are mutually exclusive. If this is the case, why not use `elif` here as well, like you do below?

extensions/payment/constants.py
rlucioni: That's fair.

rlucioni:
Just a few more minor comments. Once those are addressed and the build passes, :+1: .

PR #42, opened 2015-03-27
Using PAYMENT_PROCESSORS dict to configure payment processors

clintonb:
This implementation will allow us to more easily scale our number of payment processors without a proliferation of flat settings. Additionally, the CyberSource constructor has been updated to retrieve this configuration, so that any issues due to misconfiguration are identified early in the process.

@dianakhuang @rlucioni @stephensanchez 
FYI @feanil


extensions/settings/_oscar.py
```diff
@@ -137,9 +137,13 @@
 # PAYMENT PROCESSING
 PAYMENT_PROCESSOR_CLASS = 'extensions.payment.processors.Cybersource'
 
-CYBERSOURCE_PROFILE_ID = None
-CYBERSOURCE_ACCESS_KEY = None
-CYBERSOURCE_SECRET_KEY = None
+PAYMENT_PROCESSORS = {
```
rlucioni: What do you think of including the processor class in the processor-specific dictionaries? You could then change `PAYMENT_PROCESSOR_CLASS` to `DEFAULT_PAYMENT_PROCESSOR` and set that to `'cybersource'`.

extensions/payment/tests/test_processors.py
```diff
@@ -153,9 +157,10 @@ def _assert_order_parameters(self, order, receipt_page_url=None, cancel_page_url
             merchant_defined_data=merchant_defined_data
         )
 
+        cybersource_settings = settings.PAYMENT_PROCESSORS['cybersource']
```
rlucioni: Could you replace `'cybersource'` with `Cybersource.NAME`?

extensions/settings/_oscar.py
clintonb: @rlucioni The configuration should not know/care about code paths. You mentioned middleware when we were discussing this in person. What if we took a similar approach:

```python
# These should be set in order of preference. The first item in the list is considered the default.
PAYMENT_PROCESSORS = (
  'extensions.payment.processors.Cybersource',
  'extensions.payment.processors.Paypal',
)

# Configuration for the payment processors above
PAYMENT_PROCESSOR_CONFIG = {
  'cybersource': {...},
  'paypal': {...},
}
```

extensions/payment/tests/test_processors.py
clintonb: Done.

extensions/settings/_oscar.py
rlucioni: I like it.

clintonb: Updated.

rlucioni:
:+1:

stephensanchez:
:+1: 

I like the additional configuration options on the payment processors. What happens if we let the user choose between processors, such as PayPal vs. CyberSource (if we handled PayPal separately)? I imagine we could hard-code which processor implementation to use, but I guess the other option is to make it configurable to return the payment options from the order endpoint. Either way, additive changes, maybe not worth worrying about yet.

dianakhuang:
I know there's a couple more options in our original implementation, like url we submit the order form to. Are we just not setting those here (and instead setting them on the LMS side of things)? I just want to make sure that sort of thing is configurable.

clintonb:
@dianakhuang I added an additional entry to the config, `pay_endpoint`.

@stephensanchez We will need to update the orders endpoint to accept payment processor as input. I agree: let's worry about that problem when it arises.

extensions/settings/_oscar.py
```diff
@@ -135,11 +135,18 @@
...
 
-CYBERSOURCE_PROFILE_ID = None
-CYBERSOURCE_ACCESS_KEY = None
-CYBERSOURCE_SECRET_KEY = None
+PAYMENT_PROCESSOR_CONFIG = {
+    'cybersource': {
+        'profile_id': 'set-me-please',
+        'access_key': 'set-me-please',
+        'secret_key': 'set-me-please',
+        'pay_endpoint': 'fake-secret-key',
```
rlucioni: Minor: It might be good to change the default to a sample URL; this seems to suggest that `pay_endpoint` is a secret key.

PR #45, opened 2015-03-31
WIP: Add new endpoints for viewing the receipt page or an error page

dianakhuang:
on return from CyberSource.

@stephensanchez Here's what I have so far. We can talk it out some more later.


dianakhuang:
@rlucioni Here's the PR of the work I've been doing.

extensions/payment/constants.py
```diff
@@ -47,3 +53,88 @@ class CybersourceConstants(object):
     DECLINE = u'DECLINE'
     ACCEPT = u'ACCEPT'
     ERROR = u'ERROR'
+
+
+class CybersourceErrors(object):
+    """ The mapping of Cybersource error codes to their meanings. """
+    REASONCODE_MAP = defaultdict(lambda: "UNKNOWN REASON")
+    REASONCODE_MAP.update(
```
clintonb: Do these really need to be translated? Are they being displayed anywhere?

dianakhuang: I imagine that they'll be on the error pages we show to students eventually unless it makes more sense for us to just show them the error codes.

rlucioni:
Thanks @dianakhuang. I'll close this now and open a new PR once this is completed.

PR #46, opened 2015-03-31
Added endpoint to retrieve a user's orders

clintonb:
This endpoint supports OAuth2 access tokens, and validating said tokens against a remote provider.

@stephensanchez @rlucioni


extensions/api/authentication.py
```diff
@@ -51,3 +55,54 @@ def authenticate_credentials(self, payload):
...
+        auth = get_authorization_header(request).split()
+
+        if not auth or auth[0].lower() != b'bearer':
+            return None
+
+        if len(auth) == 1:
+            msg = _('Invalid token header. No credentials provided.')
+            raise exceptions.AuthenticationFailed(msg)
+        elif len(auth) > 2:
+            msg = _('Invalid token header. Token string should not contain spaces.')
```
rlucioni: Do these messages need to be translated? They don't seem like the kind of message that would be user-facing.

extensions/api/authentication.py
```diff
@@ -51,3 +55,54 @@ def authenticate_credentials(self, payload):
...
+            raise exceptions.AuthenticationFailed(msg)
+
+        return self.authenticate_credentials(provider_url, auth[1])
+
+    def authenticate_credentials(self, provider_url, key):
+        try:
+            response = requests.get('{}/access_token/{}/'.format(provider_url, key))
+            if response.status_code != 200:
```
rlucioni: Could you use the DRF status constant `HTTP_200_OK` instead of a naked status code?

extensions/api/authentication.py
```diff
@@ -51,3 +55,54 @@ def authenticate_credentials(self, payload):
...
+                raise exceptions.AuthenticationFailed(_('Invalid token.'))
+
+            data = response.json()
+            user = User.objects.get(username=data['username'])
+        except User.DoesNotExist:
+            raise exceptions.AuthenticationFailed(_('Invalid token.'))
+
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed(_('User inactive or deleted.'))
```
rlucioni: Similar question about these messages; do they need to be marked for translation?

extensions/api/authentication.py
```diff
@@ -51,3 +55,54 @@ def authenticate_credentials(self, payload):
...
+    Simple token based authentication.
+
+    Clients should authenticate by passing the token key in the "Authorization"
+    HTTP header, prepended with the string "Bearer ".  For example:
+
+        Authorization: Bearer 401f7ac837da42b97f613d789819ff93537bee6a
+    """
+
+    def authenticate(self, request):
+        provider_url = getattr(settings, 'OAUTH2_PROVIDER_URL', None)
```
rlucioni: Since the default value you're passing in is `None`, I would just use `get` here.
```python
provider_url = settings.get('OAUTH2_PROVIDER_URL')
```

extensions/api/tests/test_authentication.py
```diff
@@ -0,0 +1,103 @@
...
+
+from django.test import TestCase, override_settings, RequestFactory
+import httpretty
+from oscar.test import factories
+from rest_framework.exceptions import AuthenticationFailed
+
+from extensions.api.authentication import BearerAuthentication
+
+
+OAUTH2_PROVIDER_URL = 'https://example.com/oauth2'
```
rlucioni: Could you specify this in test settings?

extensions/api/tests/test_views.py
```diff
@@ -31,15 +33,36 @@
 ShippingEventType = get_model('order', 'ShippingEventType')
 
 
+class ThrottlingMixin(object):
+    def setUp(self):
+        super(ThrottlingMixin, self).setUp()
```
rlucioni: Why is it necessary to call `super()` here?

extensions/api/tests/test_views.py
```diff
@@ -370,3 +382,51 @@ def test_fulfillment_failed(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertListEqual(response.data, [])
+
+    def test_with_orders(self):
+        """ The view should return a list of the users orders, if orders exist. """
+        order = factories.create_order(user=self.user)
+        response = self.client.get(self.path, HTTP_AUTHORIZATION=self.token)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data), 1)
+        print response.data
```
rlucioni: Left over from debugging?

clintonb: Removed.

extensions/api/tests/test_authentication.py
clintonb: No. This value should only be used for these tests. This value is not necessary to run all tests.

extensions/api/authentication.py
clintonb: Updated.

extensions/api/authentication.py
clintonb: Translation removed.

extensions/api/authentication.py
clintonb: Updated.

extensions/api/authentication.py
clintonb: Translation removed.

extensions/api/tests/test_authentication.py
rlucioni: Makes sense, thanks for the clarification.

extensions/api/authentication.py
clintonb: Correction: `settings` is not a `dict`.

extensions/api/tests/test_views.py
clintonb: http://nedbatchelder.com/blog/201210/multiple_inheritance_is_hard.html

extensions/api/authentication.py
rlucioni: Right; good catch.

extensions/api/tests/test_views.py
rlucioni: Thanks for the reference.

rlucioni:
:+1:

stephensanchez:
Code looks good but I'm wondering about performance implications. I'd recommend we enforce pagination on this endpoint.

clintonb:
@stephensanchez Pagination added. If we were to add it later, technically we'd need to update our API version.

extensions/api/pagination.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import pagination
+
+
+class PageNumberPagination(pagination.PageNumberPagination):
+    """ Page number pagination with a page size of 20. """
+
+    # TODO If/when django-oscar-extensions and ecommerce are merged, set this using Django settings. See
+    # http://www.django-rest-framework.org/api-guide/pagination/#pagination.
```
rlucioni: For the time being, couldn't this be added to `_oscar.py`?

clintonb: Updated.

rlucioni:
:+1: (again).

PR #47, opened 2015-03-31
Configure Segment for use within extensions

rlucioni:
The title says it all. @clintonb or @stephensanchez, I'd appreciate a review.


clintonb:
:+1:

PR #48, opened 2015-04-01
XCOM-128: Updating the Order Serializer to include purchase info.

stephensanchez:
This PR is related to JIRA ticket: https://openedx.atlassian.net/browse/XCOM-128
Sub Task: https://openedx.atlassian.net/browse/XCOM-166
In order to get all the receipt information necessary, the Order Serializer has to be bolstered.

Adding a bit of information to the Order Serializer, including:
* billing addresses
* purchase information 
* countries

Updated unit tests to give specific validation error information.
Need additional tests that line up with the CyberSource integration. Will coordinate with @dianakhuang to make sure all purchase information is serialized properly after a completed purchase.
Tested with new receipt logic on platform.

@rlucioni @clintonb @jimabramson


extensions/api/serializers.py
```diff
@@ -3,13 +3,44 @@
...
 
 from rest_framework import serializers
+from extensions.payment.serializers import SourceSerializer
+
+
+class CountrySerializer(serializers.Serializer):
+    """Serializes a country, related to a billing address. """
+    printable_name = serializers.CharField(max_length=128)
+    #: The full official name of a country
+    #: e.g. 'United Kingdom of Great Britain and Northern Ireland'
```
rlucioni: What's up with the colons after the octothorps?

extensions/api/serializers.py
```diff
@@ -3,13 +3,44 @@
 from decimal import Decimal as D
 
 from rest_framework import serializers
+from extensions.payment.serializers import SourceSerializer
```
rlucioni: Minor: Could you separate this from the third-party import above it with a newline?

extensions/api/tests/test_integration.py
```diff
@@ -80,7 +80,8 @@ def _create_and_verify_order(self, sku):
 
         # Verify that the order data in the response is valid
         response_serializer = OrderSerializer(data=response.data)
-        self.assertTrue(response_serializer.is_valid())
+        response_serializer.is_valid()
+        self.assertEquals(response_serializer.errors, {})
```
rlucioni: Why is this preferable to checking the value of `is_valid`?

rlucioni: I think I see what you're doing: this gives better error messages on failure. Most assert methods accept an [`msg` argument](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertTrue) that, if specified, is used as the error message on failure. You could do this:
```python
self.assertTrue(response_serializer.is_valid(), msg=response_serializer.errors)
```

stephensanchez: Sure, I can do that.

extensions/api/serializers.py
stephensanchez: This comment was a cut-and-paste error when I was reviewing the model object. I'll just remove it.

rlucioni:
:+1:

extensions/api/views.py
```diff
@@ -89,7 +89,7 @@ def get_object(self):
 
         """
         order = super(RetrieveOrderView, self).get_object()
-        if order and order.is_paid and order.user == self.request.user:
+        if order and order.is_paid and order.user.username == self.request.user.username:
```
jimabramson: does this change behavior?

jimabramson:
@stephensanchez is the idea to add unit test coverage in a separate PR ?

clintonb:
@jimabramson My merge just broke your PR. Sorry about that. :(

stephensanchez:
@jimabramson yeah I'd like to add another unit test once all the PRs are in to verify we get a clean, fully serialized order with purchase information.

jimabramson:
:+1:

PR #50, opened 2015-04-02
Fix course data migration.

stephensanchez:
@rlucioni @clintonb @feanil 
The original data migration did not have the correct structure for our initial product; only a parent product should have a product class. Removing the class for the child product.
Additionally, I had a helper function for fulfillment to find the parent's product class. This is not needed. AbstractProduct already has a helper function.


rlucioni:
Looks good to me. Were you able to verify that this change allows the product's attributes (e.g., course ID) to be modified in-place?

clintonb:
:+1:

stephensanchez:
@rlucioni yep

rlucioni:
:+1:

PR #51, opened 2015-04-02
Updating version of extensions.

stephensanchez:
Updating version of extensions. Is this the only location where we specify the version?

@rlucioni


clintonb:
:+1:

rlucioni:
Yes. After merging this, you need to tag and publish a new release before [your build](https://github.com/edx/ecommerce/pull/48) over in the ecommerce will pass.

